Issue #8, opened 2020-05-29
Cannot view dashboard


cemremengu:
The link `http://localhost:8000/dashboard/overview` does not work. The content appears but quickly disappears:

![image](https://user-images.githubusercontent.com/1297759/83261591-431f2d00-a1c4-11ea-9525-9beaf1bb9f1f.png)

`http://localhost:8000/dashboard/jobs` and the rest of the tabs works fine however.

Once I enqueue some jobs, it starts to work

rdehuyss:
I don't really understand how this error happens - the only reason I can think of is that there is no BackgroundJobServer available - so not a single BackgroundJobServer was configured... 

This way, I could reproduce it. Was this perhaps the case?

cemremengu:
I will try to reproduce it and let you know

PR #10, opened 2020-05-29
Reference solution for #8

cemremengu:
This should be applied to others if you approve this approach or this can be solved at the overview page


rdehuyss:
Thanks for your contribution! I'll look into it asap (not behind the PC right now).

rdehuyss:
I fixed it by adding a condition check in the overview page mentioning that no background job servers are available. That way, it had to be fixed only in one place.

PR #58, opened 2020-10-29
Allow to change mongo database name, set jobrunr as default

jakub-hirnsal:
We can also pass some Configuration class for further configuration instead of just String name.


rdehuyss:
LGTM!

PR #76, opened 2020-12-14
Fix fail tests running at 59 minute

metlaivan:


rdehuyss:
Thx for the PR - I will never it ASAP. It's a bit busy at work now.

Issue #63, opened 2020-11-03
[Question] State of support for Kotlin/Koin


JoonasC:
Hi,
How well does JobRunr support dependency injection frameworks for Kotlin, such as [Koin](https://insert-koin.io/)?

rdehuyss:
No idea's, to be honest - from what I know, there were problems with Kotlin due to the fact that the kotlin compiler compiles its lambda's in a different way than the JVM.

Have you tried JobRunr already in Kotlin?

JoonasC:
Nope,
Haven't tried it yet.
I was just wondering whether it was worth it, to start using it.

JoonasC:
I'll try seeing if it works in Kotlin, and i'll post the results here.

JoonasC:
Ok, seems like it doesn't work.

I'm getting the following log output:
```
19:35:23.153 [main] INFO org.jobrunr.server.threadpool.ScheduledThreadPoolJobRunrExecutor - ThreadManager of type 'ScheduledThreadPool' started
19:35:23.212 [backgroundjob-zookeeper-pool-1-thread-1] INFO org.jobrunr.server.ServerZooKeeper - Server 85944396-2d08-47ba-8334-9aa94baff142 is master
19:35:23.364 [Thread-0] INFO org.jobrunr.server.BackgroundJobServer - JobRunr BackgroundJobServer (85944396-2d08-47ba-8334-9aa94baff142) and 128 BackgroundJobPerformers started successfully
19:35:23.412 [main] INFO org.jobrunr.dashboard.JobRunrDashboardWebServer - JobRunr Dashboard started at http://localhost:8000
Exception in thread "main" java.lang.IllegalArgumentException: Please provide a lambda expression (e.g. BackgroundJob.enqueue(() -> myService.doWork()) instead of an actual implementation.
	at org.jobrunr.jobs.details.SerializedLambdaConverter.toSerializedLambda(SerializedLambdaConverter.java:16)
	at org.jobrunr.jobs.details.JobDetailsAsmGenerator.toJobDetails(JobDetailsAsmGenerator.java:24)
	at org.jobrunr.scheduling.JobScheduler.enqueue(JobScheduler.java:79)
	at ApplicationKt.main(Application.kt:16)
	at ApplicationKt.main(Application.kt)
19:35:24.124 [backgroundjob-zookeeper-pool-1-thread-2] DEBUG org.jobrunr.server.JobZooKeeper - Looking for recurring jobs... 
[...]

Issue #59, opened 2020-10-29
Unable to Autowired service classes


rasi-natbuild:
**Describe the bug**

When I call my service call in try block it works fine, which means my service class is properly wired, but when I try to put it to BackgroundJob.enqueue or scheduleRecurrently it doesn't work.

**Environment**
I'm using version 1.1.0
I'm using SpringBoot 2.3.3.RELEASE
I'm running on JRE / JDK 14 
I'm using the following StorageProvider InMemoryStorageProvider

**To Reproduce**
```java
@Configuration
public class Runner {

	private static final Logger log = LoggerFactory.getLogger(Runner.class);

	@Autowired JobScheduler jobScheduler;
	@Autowired FtpAccountService ftpAccountService;

        @Bean
	public void scheduleRecurrently() {
		log.info("Configured");
		
		try {
			ftpAccountService.sweepFolders();
		} catch (Exception e) {
			e.printStackTrace();
		}
		
		BackgroundJob.enqueue(() -> ftpAccountService.sweepFolders());
	}

}
```

**Error**
```
org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'com.onlykalu.tn.edi.service.FtpAccountService' available
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:352) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:343) ~[spring-beans-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.springframework.context.support.AbstractApplicationContext.getBean(AbstractApplicationContext.java:1127) ~[spring-context-5.2.8.RELEASE.jar:5.2.8.RELEASE]
	at org.jobrunr.server.runner.BackgroundJobWithIocRunner.supports(BackgroundJobWithIocRunner.java:21) ~[jobrunr-1.1.0.jar:na]
	at org.jobrunr.server.BackgroundJobServer.lambda$getBackgroundJobRunner$0(BackgroundJobServer.java:133) ~[jobrunr-1.1.0.jar:na]
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:176) ~[na:na]
[...]

PR #78, opened 2020-12-20
Add kotlin-support module to allow enqueueing Kotlin lambda functions

### LINKED ISSUE #77: [Feature] Add support for enqueueing Kotlin functional types
Right now calling `jobScheduler.enqueue` (or `scheduleRecurrently`) with a Kotlin lambda fails since the semantics of how they are implemented internally are different from Java lambdas.

I have a fork of jobrunr with work in progress to support this use case but before I keep going down the path of making it a mergeable change, I wanted to run it by you to see if there's any interest on adding support for that.
If so, I can go ahead and continue polishing it and might be able to submit a pull request in the next few days.

Let me know what your thoughts are!

felipecsl:
Fixes #77

This is a very preliminary pass but wanted to get feedback on it before adding the bells and whistles, more test cases, etc.
Please let me know if you agree with the overall direction

Sub tasks: 

- [x] Simple function type with no bound variables
- [x] Capture bound variables
- [x] Recurring job support
- [ ] support for lambda parameters (passed to the `invoke` method)
- [x] method references


core/src/main/java/org/jobrunr/jobs/context/JobRunrDashboardLogger.java
```diff
@@ -10,7 +10,7 @@
 
 public class JobRunrDashboardLogger implements Logger {
 
-    private static final ThreadLocal<JobDashboardLogger> jobDashboardLoggerThreadLocal = new ThreadLocal<>();
+    private static JobDashboardLogger jobDashboardLogger;
```
felipecsl, line 13: this change is actually unrelated, I just had it on my branch and would be happy to revert and submit in a separate PR.
I noticed that `JobRunrDashboardLogger` was only working if called from the same thread where the job was started. Calling from a different thread would not log anything, this change should fix it

rdehuyss:
I'll look into it this evening CET.

core/src/main/java/org/jobrunr/jobs/context/JobRunrDashboardLogger.java
felipecsl: Reverted

felipecsl:
Not exactly sure why, but when i configure jobrunr with `useJobActivator(applicationContext::getBean)`, then processing the job fails with:
```
org.springframework.beans.factory.NoSuchBeanDefinitionException
org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'dog.stocks.jobrunner.StocksDogJobScheduler$scheduleJobs$2' available
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:351)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:342)
	at org.springframework.context.support.AbstractApplicationContext.getBean(AbstractApplicationContext.java:1177)
	at org.jobrunr.server.runner.BackgroundJobWithIocRunner.supports(BackgroundJobWithIocRunner.java:21)
	at org.jobrunr.server.BackgroundJobServer.lambda$getBackgroundJobRunner$0(BackgroundJobServer.java:155)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:176)
	at java.base/java.util.Spliterators$ArraySpliterator.tryAdvance(Spliterators.java:958)
	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:127)
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:502)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:488)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.findFirst(ReferencePipeline.java:543)
	at org.jobrunr.server.BackgroundJobServer.getBackgroundJobRunner(BackgroundJobServer.java:156)
	at org.jobrunr.server.BackgroundJobPerformer.runActualJob(BackgroundJobPerformer.java:77)
	at org.jobrunr.server.BackgroundJobPerformer.run(BackgroundJobPerformer.java:38)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:830)
```

core/src/main/java/org/jobrunr/utils/reflection/ReflectionUtils.java
```diff
@@ -273,10 +263,11 @@ public static void makeAccessible(AccessibleObject accessibleObject) {
     }
 
     private static <T> Constructor<T> getConstructorForArgs(Class<T> clazz, Class<?>[] args) {
-        Constructor<?>[] constructors = clazz.getConstructors();
-
+        Constructor<?>[] constructors = clazz.getDeclaredConstructors();
```
rdehuyss, line 266: What's the reason you use `getDeclaredConstructors`?

core/src/main/java/org/jobrunr/server/runner/BackgroundJobWithoutIocRunner.java
```diff
@@ -4,15 +4,14 @@
...
-import static org.jobrunr.utils.reflection.ReflectionUtils.hasDefaultNoArgConstructor;
 
 public class BackgroundJobWithoutIocRunner extends AbstractBackgroundJobRunner {
 
     @Override
     public boolean supports(Job job) {
         JobDetails jobDetails = job.getJobDetails();
         assertJobExists(jobDetails);
-        return !jobDetails.getStaticFieldName().isPresent() && hasDefaultNoArgConstructor(jobDetails.getClassName());
+        return !jobDetails.getStaticFieldName().isPresent();
```
rdehuyss, line 14: What's the reason to remove the `noArgConstuctor` requirement?

rdehuyss (commented):
Nice work! I did not know you could analyze a lambda so easy in Kotlin.

rdehuyss:
> Not exactly sure why, but when i configure jobrunr with `useJobActivator(applicationContext::getBean)`, then processing the job fails with:
> 
> ```
> org.springframework.beans.factory.NoSuchBeanDefinitionException
> org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'dog.stocks.jobrunner.StocksDogJobScheduler$scheduleJobs$2' available
> 	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:351)
> 	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:342)
> 	at org.springframework.context.support.AbstractApplicationContext.getBean(AbstractApplicationContext.java:1177)
> 	at org.jobrunr.server.runner.BackgroundJobWithIocRunner.supports(BackgroundJobWithIocRunner.java:21)
> 	at org.jobrunr.server.BackgroundJobServer.lambda$getBackgroundJobRunner$0(BackgroundJobServer.java:155)
> 	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:176)
> 	at java.base/java.util.Spliterators$ArraySpliterator.tryAdvance(Spliterators.java:958)
> 	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:127)
> 	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:502)
> 	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:488)
> 	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
> 	at java.base/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150)
> 	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
> 	at java.base/java.util.stream.ReferencePipeline.findFirst(ReferencePipeline.java:543)
> 	at org.jobrunr.server.BackgroundJobServer.getBackgroundJobRunner(BackgroundJobServer.java:156)
> 	at org.jobrunr.server.BackgroundJobPerformer.runActualJob(BackgroundJobPerformer.java:77)
> 	at org.jobrunr.server.BackgroundJobPerformer.run(BackgroundJobPerformer.java:38)
> 	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
> 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
> 	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
> 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
> 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
> 	at java.base/java.lang.Thread.run(Thread.java:830)
> ```

I had something similar before and it was caused because of different class loaders... See issue #59. Might this be the case here also?

core/src/main/java/org/jobrunr/server/runner/AbstractBackgroundJobRunner.java
```diff
@@ -50,12 +50,11 @@ protected Method getJobMethodToPerform(Class<?> jobToPerformClass) {
 
         protected void invokeJobMethod(Object jobToPerform, Method jobMethodToPerform) throws Exception {
             Object[] jobParameterValues = jobDetails.getJobParameterValues();
-
             OptionalInt indexOfJobContext = IntStream.range(0, jobDetails.getJobParameters().size())
                     .filter(i -> jobDetails.getJobParameters().get(i).getClassName().equals(JobContext.class.getName()))
                     .findFirst();
             indexOfJobContext.ifPresent(index -> jobParameterValues[index] = new RunnerJobContext(job));
-
+            jobMethodToPerform.setAccessible(true);
```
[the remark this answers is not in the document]
felipecsl: Even though the Kotlin lambda `invoke` method (the method we're calling here) has the modifier `public,` we can't call it here without calling `setAccessible(true)` first. I'm not entirely sure why (maybe the class itself is not public)? 
If you're concerned about that, we could maybe guard the call to `setAccessible` with an `if (method.name == "invoke")`?

core/src/main/java/org/jobrunr/utils/reflection/ReflectionUtils.java
felipecsl: The lambda constructor is not public :(

core/src/main/java/org/jobrunr/server/runner/BackgroundJobWithoutIocRunner.java
felipecsl: Kotlin lambda bound variables are passed into the generated class by using constructor parameters

core/src/main/java/org/jobrunr/scheduling/JobScheduler.java
```diff
@@ -292,6 +294,10 @@ public String scheduleRecurrently(JobLambda job, String cron) {
         return scheduleRecurrently(null, job, cron);
     }
 
+    public String scheduleRecurrently(String cron, Function<?> lambda) {
```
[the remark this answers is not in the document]
felipecsl: Yep! Thanks for clarifying. In Kotlin, if a lambda is the last argument of a function, it can be moved outside of the parameters, eg.:
```kotlin
jobScheduler.scheduleRecurrently("0 1 * * 1-5") {
  doSomething()
}
```
as opposed to:
```kotlin
jobScheduler.scheduleRecurrently({   doSomething() }"0 1 * * 1-5")
```
For that reason, it's a common Kotlin idiom to have a lambda as the last parameter of a function call

core/src/main/java/org/jobrunr/jobs/RecurringJob.java
```diff
@@ -62,7 +62,7 @@ public Instant getNextRun() {
     private String validateAndSetId(String input) {
         String result = Optional.ofNullable(input).orElse(getJobSignature());
 
-        if (!result.matches("[\\dA-Za-z-_(),.]+")) {
+        if (!result.matches("[\\dA-Za-z-_()\\s$,.]+")) {
```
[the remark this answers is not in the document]
felipecsl: Kotlin methods are allowed to have spaces, as you might have seen in the test cases I added, eg:
`fun `foo bar baz`() { }`
It shouldn't be a problem to use spaces in the Job ID for using in the URL as long as you [URLencode](https://docs.oracle.com/javase/7/docs/api/java/net/URLEncoder.html#encode(java.lang.String,%20java.lang.String)) it.

felipecsl:
https://github.com/jobrunr/jobrunr/pull/78#issuecomment-748671284 this actually stopped happening now.. not sure what was going on, go figure...

felipecsl:
It's working like magic now 🧙 

There's a lot of caveats though and a lot of it was just me struggling to wrap my head around how jobrunr handles job serialization and parameters. 

I'm used to relying heavily on constructor injection so it took me a while to modify my jobs to make it so they had a zero parameter constructor and no external dependencies, which made serialization fail.

A major caveat is that a lot of scenarios simply won't work and will fail miserably with a cryptic exception that will be hard for newcomers to understand.

Example (when the lambda's enclosing class doesn't have a zero-arguments default constructor:
```kotlin
class EnqueueWillFail(private val foo: Int) {
  fun someMethod() = "foo"

  fun testEnqueue() {
    // fails because EnqueueWillFail class constructor takes arguments
    jobScheduler.enqueue(::someMethod)
  }
}
```
Error:
```
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `org.jobrunr.scheduling.KotlinJobSchedulerTest$TestLambdaContainerWithArg` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
 at [Source: (String)"{"version":0,"jobSignature":"org.jobrunr.scheduling.KotlinJobSchedulerTest$test enqueue lambda with method reference$jobId$1.invoke()","jobName":"org.jobrunr.scheduling.KotlinJobSchedulerTest$test enqueue lambda with method reference$jobId$1.invoke()","jobDetails":{"className":"org.jobrunr.scheduling.KotlinJobSchedulerTest$test enqueue lambda with method reference$jobId$1","staticFieldName":null,"methodName":"invoke","jobParameters":["java.util.ArrayList",[]],"jobBoundVariables":["java.util.Arra"[truncated 437 chars]; line: 1, column: 691] (through reference chain: org.jobrunr.jobs.Job["jobDetails"]->org.jobrunr.jobs.JobDetails["jobBoundVariables"]->java.util.ArrayList[0]->org.jobrunr.jobs.JobParameter["object"])
	at com.fasterxml.jackson.databind.exc.InvalidDefinitionException.from(InvalidDefinitionException.java:67)
	at com.fasterxml.jackson.databind.DeserializationContext.reportBadDefinition(DeserializationContext.java:1615)
	at com.fasterxml.jackson.databind.DatabindContext.reportBadDefinition(DatabindContext.java:400)
	at com.fasterxml.jackson.databind.DeserializationContext.handleMissingInstantiator(DeserializationContext.java:1077)
	at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.deserializeFromObjectUsingNonDefault(BeanDeserializerBase.java:1332)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeFromObject(BeanDeserializer.java:331)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeOther(BeanDeserializer.java:199)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:166)
	at com.fasterxml.jackson.databind.jsontype.impl.AsPropertyTypeDeserializer._deserializeTypedForId(AsPropertyTypeDeserializer.java:132)
	at com.fasterxml.jackson.databind.jsontype.impl.AsPropertyTypeDeserializer.deserializeTypedFromObject(AsPropertyTypeDeserializer.java:99)
	at com.fasterxml.jackson.databind.jsontype.impl.AsPropertyTypeDeserializer.deserializeTypedFromAny(AsPropertyTypeDeserializer.java:195)
	at com.fasterxml.jackson.databind.deser.std.UntypedObjectDeserializer$Vanilla.deserializeWithType(UntypedObjectDeserializer.java:710)
	at com.fasterxml.jackson.databind.deser.impl.FieldProperty.deserializeAndSet(FieldProperty.java:147)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:293)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:156)
	at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:290)
	at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:249)
	at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:26)
	at com.fasterxml.jackson.databind.jsontype.impl.AsArrayTypeDeserializer._deserialize(AsArrayTypeDeserializer.java:120)
	at com.fasterxml.jackson.databind.jsontype.impl.AsArrayTypeDeserializer.deserializeTypedFromArray(AsArrayTypeDeserializer.java:53)
	at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserializeWithType(CollectionDeserializer.java:318)
	at com.fasterxml.jackson.databind.deser.impl.FieldProperty.deserializeAndSet(FieldProperty.java:147)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:293)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:156)
	at com.fasterxml.jackson.databind.deser.impl.FieldProperty.deserializeAndSet(FieldProperty.java:138)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:293)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:156)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4526)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3468)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3436)
	at org.jobrunr.utils.mapper.jackson.JacksonJsonMapper.deserialize(JacksonJsonMapper.java:67)
```

Anther situation that will fails is if the enclosing class has fields that cannot be serialized by Jackson (eg.: because of a reference cycle)

```
Caused by: com.fasterxml.jackson.databind.JsonMappingException: Infinite recursion (StackOverflowError) (through reference chain: com.fasterxml.jackson.databind.ObjectMapper["_jsonFactory"] (redacted...)
	at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:770)
	at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:178)
```

I'm not sure if these limitations only affect Kotlin lambdas or if Java lambdas also suffer from it (my guess is that these are Kotlin only issues), so we might want to add some safeguards to prevent users from running into these issues (eg.: throw if the enclosing class doens't have a no args constructor).
Let me know what your thoughts are!

felipecsl:
This is ready for review, I think it should be good to merge as is and we can follow up with more fixes and corner cases handling in future PRs. I'm happy with the current state

core/src/main/java/org/jobrunr/jobs/RecurringJob.java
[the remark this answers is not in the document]
rdehuyss: 👍

rdehuyss:
I'm currently reviewing it and have some feedback:
- Don't we miss the methods on the KotlinJobScheduler to schedule at certain times (the `schedule` methods)?
- Isn't it be better to separate the KotlinJobScheduler from the Java JobScheduler (perhaps via an abstract super class that shares the common methods?)
- there are 34 tests failing (because of the `jobBoundVariables` field added in `JobDetails`). I'm hoping we can do without it (by solving the problem in another way). I'll look into it next week.

rdehuyss:
I'll probably work on this tomorrow.

rdehuyss:
@felipecsl, @JoonasC: I've further looked into this:
- I prefer `JobScheduler` and `BackgroundJob` to not depend on any kotlin libs (this would otherwise require all java users to also depend on kotlin which is a no go for me)
- I wonder how DI is done (e.g. support for Koin)? Since the Kotlin jobs are created using newInstance, there is no dependency injection. As it is distributed, my guess is that it will give trouble.

I created a new branch called kotlin with:
- `JobScheduler` and `BackgroundJob` to not depend on any kotlin libs
- `KtJobScheduler` that depends on StorageProvider instead on `JobScheduler`
- I added the missing docs and missing methods (support for scheduling jobs, support for time zones)
- extra tests for scheduling jobs

Before I can merge to master, the following needs to be resolved:

- [ ] support for Koin (or any other DI framework)
- [ ] are `jobBoundVariables` really necessary? It looks like this will be resolved if we do proper DI
- [ ] all tests need to pass

All help is appreciated!

rdehuyss:
@felipecsl if you edit the PR so that maintainers can also push, I can push my changes to this PR so we keep one source of truth.

See https://code.mendhak.com/update-fork-pull-request/ for more info.

felipecsl:
image
the box is checked so you should be able to do it!

JoonasC:
@rdehuyss If you don't want all users to depend on Kotlin libs, perhaps the Kotlin support could be an addon, such as:
```
<dependency>
  <groupId>org.jobrunr</groupId>
  <artifactId>kotlin-support</artifactId>
  <version>1.0</version>
</dependency>
```

Also, I managed to get Koin to work without needing to change anything: [https://github.com/jobrunr/jobrunr/issues/63#issuecomment-731843880](https://github.com/jobrunr/jobrunr/issues/63#issuecomment-731843880).

felipecsl:
https://github.com/jobrunr/jobrunr/pull/78#issuecomment-753476429 agreed, this sounds like a good compromise to me!

JoonasC:
@rdehuyss @felipecsl Any updates on this?

felipecsl:
i havent worked on it recently but I can definitely take a look this weekend to address the feedback!

felipecsl:
@rdehuyss I looked at your branch but I'm still confused about the direction you want to take this. Let me try to enumerate all the questions:
1. "I prefer JobScheduler and BackgroundJob to not depend on any kotlin libs" - That's fair, however that means we can't enqueue Kotlin lambdas from a static context (eg.: like `BackgroundJob.enqueue()`) unless we use the same static method pattern for `KtJobScheduler` too. Is that what you had in mind?
2. "support for Koin (or any other DI framework)" - I have personally never used Koin (although I'm familiar with other DI frameworks like Dagger, for example). I'm not sure that's in scope for this PR, also unclear if jobrunr should be prescriptive about which DI framework people should use. IMHO it should be agnostic, but that's just my opinion
3. Failing tests - We can certainly fix them before merging this PR, we can do that last, I think
4. "are jobBoundVariables really necessary" - My understanding is yes, due to the way Kotlin lambdas work. Maybe I'm missing something? I'd be happy to try other alternatives if you have any

rdehuyss:
Hi @felipecsl ,

Regarding your questions:
1. My initial idea was to have the kotlin related stuff also in JobRunr Core (so not a separate jar) but not to depend/reference it from `BackgroundJob` and `JobScheduler` (the same idea as with the database support - the code is there but is only used when you use Kotlin. So, we would have a KtJobScheduler which would be referenced from projects using Kotlin. However, I also really liked the idea from @JoonasC to have a separate Kotlin addon. In that addon, we could also add a static `KtBackgroundJob` ? What do you think? 
2. Support for Koin / or any other DI - I completely agree that it should be DI agnostic. My problem is that I don't know enough about Kotlin :-).
3. Ok!
4.  I just debugged through some of the tests to understand more about Kotlin. Is it an option that if we go for the addon approach, in that addon we have a `KtBackgroundJobServer` where we would override the `private List<BackgroundJobRunner> initializeBackgroundJobRunners(JobActivator jobActivator)` (we need to make it protected off-course)? Since we then know we're always in a Kotlin runtime, all JobParameters will always be `JobBoundVariables` and we could reuse the `JobParameters` field?

Looking forward to your feedback!

felipecsl:
1. I agree! Sounds like we're in the same page then. I'll go ahead and create a separate `kotlin-support` module and move all the relevant code in there
2. Well the good news is that Kotlin classes work exactly the same way as Java classes, so you can do the same things. Constructor injection works the same way too. I don't know what you have in mind, but if you want to expand a bit more on reasons you'd need a DI framework here, we could then brainstorm ideas.
3. 👍 
4. Ohh that makes sense! I see what you're saying, just give jobParameters a new meaning, i like that approach

I think I have all the info I need to move forward now, thanks for your answers!

felipecsl:
Added `KtBackgroundJob`. One drawback of the separate module approach though, is that `JobRunrConfiguration` cannot initialize `KtBackgroundJob.jobScheduler` because that would cause a circular dependency. Users will have to manually set it with `KtBackgroundJob.jobScheduler = jobScheduler`

JoonasC:
@felipecsl Maybe the Fluent API could have an extension function which would set the job scheduler, such as:
```
JobRunr.configure()
  .useKotlin()
```

felipecsl:
ooh thats a great idea @JoonasC 👍 thank you

rdehuyss:
I really like where this is heading - I'm now looking into how we can avoid the jobBoundVariables.

Thanks for all the work, @felipecsl !

rdehuyss:
I merged the PR again in the kotlin branch and added one more test - how to use DI in jobs.

When running that test, it did not work as I expected... 

@felipecsl What do you think? I think there is something wrong with the lambda analysis:
See https://github.com/jobrunr/jobrunr/blob/kotlin/kotlin-support/src/test/kotlin/org/jobrunr/scheduling/KtJobSchedulerTest.kt#L95

Shouldn't the `JobDetails.className` be the FQN of TestService? Because as it is now, it will not be loaded from the IoC container.

JoonasC:
Sorry to pester you on this, but any updates @rdehuyss @felipecsl?

thais:
I'm rooting for this pull request to be merged.

felipecsl:
we've been slow to make progress on this, but will pick it up again soon!

rdehuyss:
I've looked into this and think I may have found a way. It's totally different than @felipecsl approach and there may be better ways to do so...
It involves bytecode analysis of the kotlin classes, the same way I do for the Java classes.

rdehuyss:
FYI for those that are interested: https://github.com/jobrunr/jobrunr/tree/2.0...

I think I may have an RC within a week or two.

felipecsl:
@rdehuyss regarding https://github.com/jobrunr/jobrunr/pull/78#issuecomment-773823360
What's the advantage of analyzing the bytecode instead of using reflection? any pros/cons to that approach?

rdehuyss:
Hi @felipecsl , no advantages - but it's the only way I've found to correctly extract the necessary information.

In Java, there is no other way to find out what method was called inside a Java lambda.
In kotlin, I'm not sure but I also think that is the case. See also https://github.com/jobrunr/jobrunr/pull/78#issuecomment-759489940 - I did not find a way to find the reference to TestService which is needed for DI.

rdehuyss:
Calling out to all kotlin developers: 

## Stream vs Sequence
In Java you have a Stream of objects (for example returned by Spring Data). This stream is better as it does not take all items in memory (whereas a `java.util.List` does).

Is the equivalent of that in Kotlin a Sequence? 

Let me rephrase my question: how are items returned from your database if you do not want to take them all into memory? E.g., to iterate all users in your database: is it a `Sequence<User>`?

## Generics in Kotlin
In Java, we can use the following construct to call the method `MyService.doWork()` in a BackgroundJob. This comes in handy when you do not have an instance of MyService available.
`UUID jobId = BackgroundJob.<MyService>enqueue(x -> x.doWork());`

What is the equivalent Kotlin syntax?

JoonasC:
@rdehuyss On generics, I'm not sure whether this exactly fits your use case but it seems the syntax is:
```
BackgroundJob.enqueue<MyService> {
  it.doWork()
}
```

Reference: [https://pl.kotl.in/SmBVlfHYG](https://pl.kotl.in/SmBVlfHYG)

rdehuyss:
For the brave kotlin developers that want to try out JobRunr 2.0... please read issue https://github.com/jobrunr/jobrunr/issues/77 and report back there.

I'll close this PR for now as the approach used to add Kotlin support is different from the original approach.

@felipecsl: thanks for your PR - I could not have done this without you setting up the project structure!

felipecsl:
@rdehuyss regarding https://github.com/jobrunr/jobrunr/pull/78#issuecomment-776531118
Yes, the Kotlin equivalent of a Java `Stream` is `Sequence`

PR #98, opened 2021-02-23
Allow configure dashboard path

jakub-hirnsal:


rdehuyss:
Hi @jakub-hirnsal , did you try this locally does it work for you? I think that the React Frontend also contains a lot of references to "/dashboard" so I think it requires some more work.

jakub-hirnsal:
Oh, I see. I haven't tried it locally.

rdehuyss:
Since this PR doesn't work, I'll close it for now.

Feel free to create a new PR where the change is tested locally.

PR #103, opened 2021-03-04
Added koin job activator

JoonasC:
Adds a job activator for [Koin](https://insert-koin.io/).

### DIFF
--- a/jobrunr-kotlin-support/build.gradle
+++ b/jobrunr-kotlin-support/build.gradle
@@ -7,6 +7,8 @@ plugins {
 dependencies {
   api project(':core')
 
+  api 'org.koin:koin-core:2.2.2'
+
   testImplementation project(':core')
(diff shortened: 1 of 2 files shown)


jobrunr-kotlin-support/build.gradle
```diff
@@ -7,6 +7,8 @@ plugins {
 dependencies {
   api project(':core')
 
+  api 'org.koin:koin-core:2.2.2'
```
rdehuyss, line 10: Would this also work with a compileOnly dependency? That way, if users don't use Koin, they don't get this transitive dependency?

JoonasC: Sorry, for some reason I didn't get the email notification that you had reviewed the PR. I think in that case it would maybe be better to split the koin job activator into a separate koin support plugin?

PR #113, opened 2021-05-05
Fix logging

benallard:
As a side-note, there seems to be a missing entry in the best-practice page about that.

(here I call a getter to an injected bean. How should I improve that ?)

### DIFF
--- a/core/src/main/java/org/jobrunr/jobs/details/instructions/JobDetailsInstruction.java
+++ b/core/src/main/java/org/jobrunr/jobs/details/instructions/JobDetailsInstruction.java
@@ -41,7 +41,7 @@ public class JobDetailsInstruction extends VisitMethodInstruction {
             final Object result = getObject();
             final long after = System.nanoTime();
             if ((after - before) > 1_000_000) {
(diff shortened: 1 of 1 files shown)


rdehuyss:
Thx! That one I missed.

Regarding your side-note: Can you start a discussion about that as I cannot completely follow what you mean?

benallard:
Sure, where’s the best place to do that?

I simply meant that the logs refer me to the “best-practice” page, but I cannot find a matching entry relating to my issue on that page.

rdehuyss:
Hi @benallard - the best place to do so is here: [Github discussions](https://github.com/jobrunr/jobrunr/discussions/categories/q-a)

If you create a question with a code example (your job with the getter to injected bean), I hopefully have a bit more context to understand what is happening and I can then share my idea.

I hope that the Github discussions will become kind of a place where some FAQ's are answered so that developers can quickly find answers to some common questions.

benallard:
I didn't realised 'Discussion' were a thing. might have skipped the memo from github about that. They hooked us up on talking about PR everywhere, good luck about marginalizing that term as well :).

rdehuyss:
:rofl:

PR #174, opened 2021-07-17
Rename schemaName to tablePrefix.

tan9:


rdehuyss:
Thanks for this - I thought I did all the renames but must have overseen these for some reason.

PR #180, opened 2021-07-22
Add back deprecated method in public configuration for compatibility.

tan9:
I think it is good to retain background compatibility while evolving.

Renaming a  public method directly will cause a `MAJOR` version bump in case of following [Semantic Versioning](https://semver.org/#semantic-versioning-200), and I don't think people want to upgrade to a new version of a library just for the more reasonable naming of things :smile:


rdehuyss:
Hi @tan9 ,

Thanks for this... However, the next release will be 4.0 as I'm working to make JobRunr integrate better with Spring, Quarkus and Micronaut.

The other reason for a 4.0 release is that it will then be covered by Infoq... So the release will take a bit longer this time.

tan9:
@rdehuyss  I got it, looking forward to seeing the 4.0 release :smile:

PR #181, opened 2021-07-23
interacting with popular third-party application monitoring systems.

neildeng:
I propose to export some metrics, like jobs counter, background job server status, etc. 
So jobrunr can be monitor on popular third-party application monitoring systems, such as Prometheus, Datadog, New Relic, WaveFront, dynaTrace, etc.

### DIFF
--- a/jobrunr-spring-boot-starter/src/main/java/org/jobrunr/metric/StorageProviderMetricsBinder.java
+++ b/jobrunr-spring-boot-starter/src/main/java/org/jobrunr/metric/StorageProviderMetricsBinder.java
@@ -1,21 +1,24 @@
 package org.jobrunr.metric;
 
-import io.micrometer.core.instrument.Gauge;
-import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.Tags;
 import org.jobrunr.jobs.states.StateName;
-import org.jobrunr.storage.PageRequest;
+import org.jobrunr.storage.JobStats;
 import org.jobrunr.storage.StorageProvider;
+import org.jobrunr.storage.listeners.JobStatsChangeListener;
 
 import javax.annotation.PostConstruct;
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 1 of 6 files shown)


jobrunr-spring-boot-starter/src/main/java/org/jobrunr/metric/StorageProviderMetricsBinder.java
```diff
@@ -0,0 +1,35 @@
...
+        this.registryStorageProvider();
+    }
+
+    private void registryStorageProvider() {
+        String PREFIX = "jobrunr_";
+        String jobs = PREFIX + "jobs";
+        Arrays.stream(StateName.values())
+                .forEach((state) -> Gauge
+                        .builder(jobs, this.storageProvider, (provider) -> (double) provider.getJobPage(state, PageRequest.ascOnUpdatedAt(0)).getTotal())
+                        .tag("state", state.toString()).register(this.registry));
```
rdehuyss, line 33: I would prefer to use provider.getJobStats() here as it will make a big difference performance wise (one DB call to a cached view vs 7 database calls).  Is there a reason why you don't use it now?

rdehuyss (changes requested):
Wauw, thanks for this, I like it a lot!

I would like to ask for one change - to use the storageProvider.getJobStats(). Or perhaps I don't see a problem you have encountered with it?

jobrunr-spring-boot-starter/src/main/java/org/jobrunr/metric/StorageProviderMetricsBinder.java
neildeng: I missed it, so I fix it and add some test cases

rdehuyss:
Hi @neildeng ,

Thanks for this, it's merged. Do note that I did some changes:
- do not create a SimpleMeterRegistry but have it injected as it differs depending on which type of registry you use (elastic, Prometheus)
- follow best practices regarding meaning

Cheers!

rdehuyss:
Hi @neildeng , can you test this with JobRunr 4.0.0-RC1? It should soon be available in Maven Central.

PR #183, opened 2021-07-27
Make `scheduleRecurrently` public

mendess:
The `JobDetails` constructor is already public and sometimes it's easier to create an instance of it than a lambda (for example when using reflection/annotations)

### DIFF
--- a/core/src/main/java/org/jobrunr/scheduling/JobScheduler.java
+++ b/core/src/main/java/org/jobrunr/scheduling/JobScheduler.java
@@ -591,7 +591,7 @@ public class JobScheduler {
         return saveJob(new Job(id, jobDetails, new ScheduledState(scheduleAt)));
     }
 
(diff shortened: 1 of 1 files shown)


rdehuyss:
Is this still relevant when the new release which has the `@Recurring` annotation is released? 

For me, it will unbalance the API as schedule and enqueue are also still package private,

mendess:
Maybe those two could be public too. I'm implementing my version of the `Recurring` annotation which has different functionality that is relevant to our usecases but probably not desirable to have in the library and having this be public would allow us and others to develop our own custom implementation.

rdehuyss:
Well, that's why they are package private (and also how I do this in the `spring-boot-starter`). Can't you add your `Recurring` annotation processor in the same package as the `JobScheduler`.

The reason I'm really hesitant on this is because of code completion in IDE's and the confusion it will create with developers.

What is your `Recurring` annotation doing that's so special? I perhaps prefer to add more features to the `Recurring` than to open up the package private part of the API.

PR #190, opened 2021-08-11
Add disabled cron exporession support.

tan9:
Delete recurring job if `@Recurring.cron` evaluated as `-`.


rdehuyss:
Hi @tan9 , I just released JobRunr 4.0.0-RC1. It should soon be available in Maven Central. Can you test and report back?

Issue #219, opened 2021-09-08
[Feature] Honor `org.jobrunr.database.tablePrefix` property value for MongoDB collections


sa1nt:
**Is your feature request related to a problem? Please describe.**

When running on [cloudfoundry](https://www.cloudfoundry.org/), the MongoDB service there doesn't allow to create Databases, and by default JobRunr MongoDBStorageProvider tries to create one to store its Mongo collections. 

While it's possible to explicitly specifiy Mongo Database to use in MongoDBStorageProvider, I was expecting that setting the `org.jobrunr.database.tablePrefix` value would make the collections created have this prefix when using JobRunr Spring Boot Starter. 
E.g. if I set `org.jobrunr.database.tablePrefix = "jobrunr_"`, I'd expect JobRunr collection names to be `jobrunr_jobs`, `jobrunr_recurring_jobs` etc.

**Describe the solution you'd like**

Make `org.jobrunr.database.tablePrefix` Spring Boot Starter property value apply to MongoDB collections.

**Describe alternatives you've considered**

That is by no means a _required_ feature (for me), so the easiest alternative would be to just keep things as they are now.

**Additional context**

None.

rdehuyss:
Resolved

PR #220, opened 2021-09-08
#219 Now honoring org.jobrunr.database.tablePrefix when using MongoDB 

sa1nt:
Feature #219


rdehuyss:
Thanks, good idea. I did change the implementation to have it consistent with all other `StorageProviders` and also made sure it works for Quarkus and Micronaut

PR #245, opened 2021-10-27
Increase log level for job exception after maxAmountOfRetriesReached

tedliang:
Increase log level for job exception from WARN to ERROR when there is no more auto retry.

### DIFF
--- a/core/src/main/java/org/jobrunr/server/BackgroundJobPerformer.java
+++ b/core/src/main/java/org/jobrunr/server/BackgroundJobPerformer.java
@@ -106,6 +106,11 @@ public class BackgroundJobPerformer implements Runnable {
         try {
             job.failed(message, e);
             saveAndRunStateRelatedJobFilters(job);
+            if (job.getState() == FAILED) {
+                LOGGER.error("Job(id={}, jobName='{}') processing failed: {}", job.getId(), job.getJobName(), message, e);
+            } else {
+                LOGGER.warn("Job(id={}, jobName='{}') processing failed: {}", job.getId(), job.getJobName(), message, e);
+            }
         } catch (IllegalJobStateChangeException ex) {
             if (ex.getFrom() == DELETED) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jobrunr/server/BackgroundJobPerformer.java
```diff
@@ -115,6 +115,12 @@ private void updateJobStateToFailedAndRunJobFilters(String message, Exception e)
         } catch (Exception badException) {
             LOGGER.error("ERROR - could not update job(id={}, jobName='{}') to FAILED state", job.getId(), job.getJobName(), badException);
         }
+
```
rdehuyss, line 118: Is there a reason you put it here? In my eyes, this block should be in the `try` statement, right before the `catch` clause.

rdehuyss (changes requested):
Hi, first of all - sorry for the late reply and thanks for this improvement!

Only one comment and great that you added the test!

core/src/main/java/org/jobrunr/server/BackgroundJobPerformer.java
tedliang: Thanks for your review. It depends on whether or not the job exception need to be logged if an error occurs in the `try` block.

I'm ok to move the log to right before the `catch` clause if you prefer not to log the job exception.

rdehuyss: Yeah, I do prefer that from a code style perspective.

This other 2 methods are JobRunr internal and normally don't fail (except e.g. when storage is down) in which case we could not update the job anyway and it will be retried.

Can you move it? Then I'll approve it and create a new release...

Thanks for all your hard work!

tedliang: No worries. Pushed the updates.

PR #269, opened 2021-11-25
correct jobrunr version on README

theexiile1305:


rdehuyss:
Thx for this! But... I'm going to close it :smile: .To make changes to the readme, it is important to have [SKIP CI] in the commit as it otherwise runs an extra build without reason. I'll update the readme though.

Issue #231, opened 2021-10-08
[Feature] Recurring Jobs with a fixed Interval


eifinger:
**Is your feature request related to a problem? Please describe.**
Hi and thank you for this project!

I have the following use case:

1. A customer registers to a program where he needs to send data 3 times per X days
2. After X days I want to check if the customer has indeed sent data 3 times

So when the customer registers I want to schedule a job "check now in X days". It has to be the same time every X days (down to the minute)

**Describe the solution you'd like**
A great api would be:

```java
BackgroundJob.scheduleRecurringly("my-recurring-job", Duration.parse("PT5D"), () -> service.doWork());
```

**Describe alternatives you've considered**
I have considered to compute a cron schedule depending on `now`.
But this leads to problems at the end of a month as described [here](https://stackoverflow.com/a/55087822)
The same is the case for a delays of hours and the end of a day.

**Additional context**
-

rdehuyss:
Solved in v5.0.0 branch.

PR #285, opened 2021-12-14
Feature/support durations in recurring jobs

daniela-tumbraegel:
Implements issue #231


rdehuyss:
Hi Daniela,

thanks for this - I'll try to review it ASAP.

I'll keep you posted!

core/src/main/java/org/jobrunr/scheduling/ScheduleFactory.java
```diff
@@ -0,0 +1,24 @@
...
+            return CronExpression.create(scheduleExpression);
+        }
+        else if(type.equals(ScheduleExpressionType.INTERVAL)){
+            return new Interval(scheduleExpression);
+        }
+        else {
+            throw new InvalidScheduleException("Schedule expression cannot be mapped to any type");
+        }
+    }
+}
```
rdehuyss, line 24: Let's merge this with the `ScheduleExpressionType`.

The enum can also have a method `create` which returns then the correct `Schedule`.

core/src/main/java/org/jobrunr/scheduling/exceptions/InvalidScheduleException.java
```diff
@@ -0,0 +1,8 @@
+package org.jobrunr.scheduling.exceptions;
+
+public class InvalidScheduleException extends RuntimeException {
+
+    public InvalidScheduleException(String message) {
+        super(message);
+    }
```
rdehuyss, line 7: Let's move this to the `Interval` package and make it a bit more smarter: have the message ("Schedule expression cannot be mapped to any type") here as it is only called from one place. Also add the actual expression the user entered and add it to the message - this makes debugging easier.

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/recurring-jobs/recurring-jobs.js
```diff
@@ -175,7 +175,7 @@ const RecurringJobs = (props) => {
                                                         {recurringJob.jobName}
                                                     </TableCell>
                                                     <TableCell>
-                                                        {cronstrue.toString(recurringJob.cronExpression)}
+                                                        {cronstrue.toString(recurringJob.scheduleExpression)}
```
rdehuyss, line 178: This will fail for durations. We could check if the scheduleExpression starts with 'P' and depending on that show a different value.

core/src/test/java/org/jobrunr/scheduling/interval/IntervalTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+import static java.time.ZoneOffset.UTC;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+
+class IntervalTest {
+
+    private static final DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+
+    private static final String TEN_SECONDS = "PT1S";
```
rdehuyss, line 23: Should be PT10S.

If I change this to PT10S, no tests fail which I would expect them to do. 

Can you also have an expectedResult in the startInstantDurationAndResultInstant and compare the expected value to the actual value? That it is only after the current instant will not prevent me from making mistakes in the future if something changes.

core/src/testFixtures/java/org/jobrunr/jobs/RecurringJobAssert.java
```diff
@@ -18,8 +18,8 @@ public RecurringJobAssert hasJobName(String name) {
         return this;
     }
 
-    public RecurringJobAssert hasCronExpression(String cronExpression) {
-        Assertions.assertThat(actual.getCronExpression()).isEqualTo(cronExpression);
+    public RecurringJobAssert hasScheduleExpresssion(String scheduleExpression) {
+        Assertions.assertThat(actual.getScheduleExpression()).isEqualTo(scheduleExpression);
```
rdehuyss, line 22: Type: `hasScheduleExpresssion` should only have two s.

rdehuyss (changes requested):
Hi Daniela,

great work on this - I don't think I could have done it better myself. 

Some feedback:
- this will be a breaking change so it will only be included for JobRunr 5.0.0 (I expect to release this somewhere end of January - February). Originally, I though about adding an extra `RecurringIntervalJob` which extends `RecurringJob` but this would make deserialization difficult.
- You're not using the standard Java formatting rules, could you apply them?
- For the rest, there is some feedback - could you fix this?

Thanks!
Ronald

daniela-tumbraegel:
Hi Ronald,

thanks for the quick review and the feedback! :) 

I'll go over the files again and auto format them using the IntelliJ default code style scheme. Hope that works for you.
I'm also going to incorporate your suggestions sometime this or next week.

Thanks again,
Daniela

core/src/test/java/org/jobrunr/scheduling/interval/IntervalTest.java
daniela-tumbraegel: Since `Instant.now()` is used here, the only way to get the expected result would be to calculate it using the same logic that is supposed to be verified by this test case. I'm not entirely sure if that would be a good idea. 

I added a new test instead that uses Instant.now() as base date and checks if the next instant is ten seconds later.

Let me know what you think :)

core/src/main/java/org/jobrunr/scheduling/ScheduleFactory.java
daniela-tumbraegel: I moved it to the `ScheduleExpressionType` class but isn't it a little odd to have a method inside an enum return an instance of `Schedule`? Any thoughts?

core/src/main/java/org/jobrunr/scheduling/ScheduleExpressionType.java
```diff
@@ -0,0 +1,36 @@
...
+        ScheduleExpressionType type = ScheduleExpressionType.getScheduleType(scheduleExpression);
+
+        if (type.equals(ScheduleExpressionType.CRON_EXPRESSION)) {
+            return CronExpression.create(scheduleExpression);
+        } else if (type.equals(ScheduleExpressionType.INTERVAL)) {
+            return new Interval(scheduleExpression);
+        } else {
+            throw new InvalidScheduleException(scheduleExpression);
+        }
+    }
```
rdehuyss, line 35: I was more thinking something in the line of:

http://tutorials.jenkov.com/java/enums.html#enum-abstract-methods.

I also think we can get away with only invalid if it is null or empty. So I would remove unknown for now. So, this would become (pseudo code, not tested): 

```java
public enum ScheduleExpressionType {
    CRON_EXPRESSION {
        @Override
        public Schedule createSchedule() {
            return CronExpression.create(scheduleExpression);
        }
    },
    INTERVAL {
        @Override
        public Schedule createSchedule() {
            return new Interval(scheduleExpression);
        }
    },
    INVALID {
        @Override
        public Schedule createSchedule() {
            throw new InvalidScheduleException(scheduleExpression);
        }
    };

    public static ScheduleExpressionType getScheduleType(String scheduleExpression) {
        if (scheduleExpression.matches(".*\\s.*")) {
            return CRON_EXPRESSION;
        } else if (scheduleExpression.startsWith("P")) {
            return INTERVAL;
        }
        return INVALID;
    }

  public abstract Schedule createSchedule(String scheduleExpression);
}
```

core/src/testFixtures/java/org/jobrunr/jobs/RecurringJobTestBuilder.java
```diff
@@ -71,7 +76,12 @@ public RecurringJobTestBuilder withJobDetails(JobDetails jobDetails) {
...
     public RecurringJobTestBuilder withCronExpression(String cronExpression) {
-        this.cronExpression = CronExpression.create(cronExpression);
+        this.schedule = CronExpression.create(cronExpression);
+        return this;
+    }
+
+    public RecurringJobTestBuilder withIntervalExpression(String intervalExpression) {
+        this.schedule = new Interval(Duration.parse(intervalExpression));
         return this;
     }
```
rdehuyss, line 86: Shouldn't we set `createdAt` here immediately? If it's a `RecurringJob` with an IntervalExpression, to be valid, it needs the `createdAt`. What I would do:

```java
    public RecurringJobTestBuilder withIntervalExpression(String intervalExpression) {
        this.withIntervalExpression(intervalExpression, Instant.now());
        return this;
    }

    public RecurringJobTestBuilder withIntervalExpression(String intervalExpression, Instant createdAt) {
        this.schedule = new Interval(Duration.parse(intervalExpression));
        this.createdAt = createdAt;
        return this;
    }
```

core/src/main/java/org/jobrunr/scheduling/interval/Interval.java
```diff
@@ -0,0 +1,70 @@
...
+        this.duration = Duration.parse(durationExpression);
+    }
+
+    @Override
+    public Instant next(Instant baseInstant, ZoneId zoneId) {
+        Duration durationUntilNow = Duration.between(baseInstant, Instant.now());
+        long amountOfDurationsUntilNow = durationUntilNow.toNanos() / duration.toNanos();
+
+        return baseInstant.plusNanos(duration.toNanos() * (amountOfDurationsUntilNow + 1));
+    }
```
rdehuyss, line 27: Let's try this for testing:

```java

    @Override
    public Instant next(Instant baseInstant, ZoneId zoneId) {
        return next(baseInstant, Instant.now(), zoneId);
    }

    @VisibleFor("testing")
    Instant next(Instant baseInstant, Instant currentInstant, ZoneId zoneId) {
        Duration durationUntilNow = Duration.between(baseInstant, currentInstant);
        long amountOfDurationsUntilNow = durationUntilNow.toNanos() / duration.toNanos();

        return baseInstant.plusNanos(duration.toNanos() * (amountOfDurationsUntilNow + 1));
    }

```

Now, you can use the `next(Instant baseInstant, Instant currentInstant, ZoneId zoneId)` in the IntervalTest. The reason I'm so picky about this is that in my eyes, otherwise the `@ParameterizedTest @MethodSource("startInstantDurationAndResultInstant")` test does not add much value but I do like it. Now, we can really test this method to the full extend of its responsibility.

core/src/test/java/org/jobrunr/scheduling/interval/IntervalTest.java
```diff
@@ -0,0 +1,119 @@
...
+        Interval interval = new Interval(TEN_SECONDS);
+        Duration duration = Duration.parse(TEN_SECONDS);
+        Instant nextInstant = interval.next(now, UTC);
+
+        assertThat(nextInstant)
+                .describedAs("Expecting %s to be after or equal to %s for duration %s and start with now()", nextInstant, now, duration)
+                .isAfterOrEqualTo(now);
+        assertThat(nextInstant)
+                .isEqualTo(now.plus(duration));
+    }
```
rdehuyss, line 61: This can be removed again.

core/src/test/java/org/jobrunr/scheduling/interval/IntervalTest.java
```diff
@@ -0,0 +1,119 @@
...
+                    .describedAs("Expecting %s to be after or equal to %s for duration %s and start date %s", nextInstant, now, duration, inputInstant)
+                    .isAfterOrEqualTo(now);
+            assertThat(nextInstant)
+                    .describedAs("Expecting %s to be before to %s for duration %s and start date %s", nextInstant, now.plus(duration), duration, inputInstant)
+                    .isBefore(now.plus(duration));
+        } catch (Exception e) {
+            System.out.printf("Error for %s and %s%n", baseDate, durationExpression);
+            throw e;
+        }
+    }
```
rdehuyss, line 47: This test can now accept to more arguments:
`arguments("2019-01-01 00:00:00", TEN_SECONDS, "2019-01-5 00:00:01", "2019-01-5 00:00:10"),`

This would thus be:
"2019-01-01 00:00:00" => baseDate
TEN_SECONDS => intervalExpression
"2019-01-5 00:00:00" => the moment when JobRunr is evaluating the expression (Instant.now() in production code)
"2019-01-5 00:00:10" => the expected outcome

Now, we can also really test corner cases to the exact moment.

rdehuyss (changes requested):
Thanks for the quick fixes, almost there.

Just some minor things including the test. Sorry I'm so picky about it but I prefer these things to be very well tested - it's me that will need to fix it if there is an issue reported by other users 😄 .

core/src/main/java/org/jobrunr/scheduling/ScheduleExpressionType.java
daniela-tumbraegel: Oh, I see! I hope the new changes are more in line of what you were thinking then ^^

rdehuyss:
Ready to merge. Will keep it open until 5.0 is ready to be released.

kibertoad:
Probably should not be closed.

rdehuyss:
Hi @daniela-tumbraegel , 

I'm afraid I will only be able to release this by the end of February / somewhere in March. I'm really sorry and want to thank you for this wonderful PR. I do commit myself to release it in Q1 this year.

The reason is that this will be part of JobRunr 5.0 and I want to finish some other stuff also for that release. 

Thanks again,
Ronald

rdehuyss:
Merged to master 😄

PR #290, opened 2021-12-19
Feature/spring boot native

rdehuyss:


rdehuyss:
Keep open ..

core/src/main/java/org/jobrunr/utils/ClassPathUtils.java
```diff
@@ -68,6 +68,8 @@ private static Path toPath(URL url) {
             }
             if ("jar".equals(uri.getScheme())) {
                 return JarFileSystemUtils.toPath(uri);
+            } else if ("resource".equals(uri.getScheme())) {
```
kibertoad, line 71: why not just use switch?

core/src/main/java/org/jobrunr/utils/ResourceFileSystemUtils.java
```diff
@@ -0,0 +1,39 @@
...
+            FileSystem fs = getFileSystem();
+            return fs.getPath(StringUtils.substringAfter(uri.toString(), ":"));
+        } catch (ProviderNotFoundException e) {
+            throw new ProviderNotFoundException("Provider not found for URI " + uri.toString());
+        }
+    }
+
+    private static FileSystem getFileSystem() throws IOException {
+        if (fileSystem == null) {
+            HashMap<String, Object> options = new HashMap<>();
```
kibertoad, line 33: `Map` should be sufficient for the left side

settings.gradle
```diff
@@ -7,8 +15,9 @@ include ':language-support:jobrunr-kotlin-16-support'
 include ':framework-support:jobrunr-micronaut-feature'
 include ':framework-support:jobrunr-quarkus-extension:deployment'
 include ':framework-support:jobrunr-quarkus-extension:runtime'
-include ':framework-support:jobrunr-quarkus-extension:tests'
+//include ':framework-support:jobrunr-quarkus-extension:tests'
```
kibertoad, line 18: should this be commented out?

rdehuyss:
Merged in feature/v5 with resolved feedback.

PR #418, opened 2022-04-13
Fix for multiple-datasource issue.

kfowler:
This commit fixes an issue with jobrunr-micronaut-feature when used with multiple, non-default data sources.

When a micronaut/jobrunr application is configured with multiple datasources--none of which are the default data source--micronaut will throw:

NonUniqueBeanException: Multiple possible bean candidates found: [javax.sql.DataSource, javax.sql.DataSource]

The error is thrown from the .orElse() on line 30 of JobRunrSqlStorageProviderFactory.java because there are multiple data sources defined when beanContext.getBean(DataSource.class) is evaluated.

The fix is to only evaluate beanContext.getBean(...) when the bean hasn't already been resolved, i.e., change .orElse() to
.orElseGet(()->...)

Example code demonstrating the issue:
https://github.com/kfowler/example-micronaut/commit/607c6681c0baf0e88967134802729ec23136a2c7
https://github.com/kfowler/example-micronaut/tree/kf/multiple-ds-error

The example code starts-up and works as-expected with this fix.


rdehuyss:
Thanks a lot!

PR #420, opened 2022-04-16
Clarify "JobRunr only supports enqueueing/scheduling of one method" error message

ermiyaeskandary:
I've unfortunately just spent upwards of an hour trying to figure out why 1 of my jobs started failing after adding in `LOGGER.INFO(...)` statements within the job.

The error message **JobRunr only supports enqueueing/scheduling of one method** makes it sound as though you can only have one method invocation.

This is not true.

For example, this is completely valid with 4 method invocations:

```java
jobScheduler.enqueue(() -> {
    final String a = serviceA.call();
    final String b = serviceB.call();
    final String c = serviceC.call();

    serviceD.doSomething(a, b, c);
});
```

What the error actually means is that, this is invalid - a void invocation which is not the last statement:

```java
jobScheduler.enqueue(() -> {
    serviceA.voidMethod()
    final String a = serviceA.call();
});
```

And this is also invalid - 2 void invocations:

```java
jobScheduler.enqueue(() -> {
    serviceA.voidMethod()
    serviceB.voidMethod()
});
```

I've updated the exception message to be much clearer regarding this expectation as currently, as it stands, it can be quite ambiguous & misleading.

I'd appreciate any and all feedback, thank you in advance.

EE

P.S. the wording on this [documentation page](https://www.jobrunr.io/en/documentation/background-methods/) ("Only one method to enqueue is supported") could also do with being updated to reflex the above i.e. something like "Only one void method to enqueue is supported (as the last statement)". I haven't dug into how/if I can update that but happy to do so as part of this MR if the docs are in this codebase.

### DIFF
--- a/core/src/main/java/org/jobrunr/jobs/details/instructions/JobDetailsInstruction.java
+++ b/core/src/main/java/org/jobrunr/jobs/details/instructions/JobDetailsInstruction.java
@@ -29,7 +29,7 @@ public class JobDetailsInstruction extends VisitMethodInstruction {
     @Override
     public Object invokeInstruction() {
         if (!isLastInstruction() && isVoidInstruction()) {
-            throw new JobRunrException("JobRunr only supports enqueueing/scheduling of one method");
+            throw new JobRunrException("JobRunr only supports enqueueing/scheduling of one void method (which also must be the last statement executed)");
         } else if (isLastInstruction()) {
             jobDetailsBuilder.setClassName(getClassName());
             jobDetailsBuilder.setMethodName(getMethodName());


rdehuyss:
Hi Ermiya,

well - only the last part of the lambda will be actually done inside a job... 

Meaning in the case you mentioned:
```java
jobScheduler.enqueue(() -> {
    final String a = serviceA.call();
    final String b = serviceB.call();
    final String c = serviceC.call();

    serviceD.doSomething(a, b, c);
});
```
The first 3 methods will be called during creating of your job - only the last method will be actually done inside a job. That's why I encourage to create the lambda's as small as possible.

I hope this clarifies it?

ermiyaeskandary:
Ahhhh, I wasn't aware - nothing actually flags this up & there's no warning to state this.

I thought all would be ran inside the Lambda!

Would using a `JobRequest` mean that they would all run together as part of the job?

ermiyaeskandary:
Actually thinking of it - would this work?

```java
jobScheduler.enqueue(() -> {
  triggerIt();
});
...

public void triggerIt() {
    final String a = serviceA.call();
    final String b = serviceB.call();
    final String c = serviceC.call();

    serviceD.doSomething(a, b, c);
}
```

rdehuyss:
Yes (using Job request) and yes (triggerIt).

The technical limitation (among others) is indeed that all of this is serialized as Json.

The best practices page also explains some of this.

ermiyaeskandary:
@rdehuyss Thank you for the confirmation!

And am I correct in that `triggerIt` must be public? No compile-time errors are thrown if the method is private..

That said, would the error `JobRunr only supports enqueueing/scheduling of one void method` still not be better? If the error had mentioned void, I would've realised that all of the logic should probably go into an encapsulating method 🙂

rdehuyss:
I'm looking into another way of doing this but it will be a big change... it will be for JobRunr 6 or 7.

Feel free to try but then you get in the nitty gritty details of JVM bytecode.

And, it does not need to be a void method by the way, you can return something but nothing will be done with the result in the JobRunr OSS version.

PR #478, opened 2022-06-01
JobRunr page request size configuration. Fixes #477

### LINKED ISSUE #477: [Feature] Allow configuration of page request sizes used by JobZooKeeper
**Is your feature request related to a problem? Please describe.**
Currently, the hardcoded values used by JobZooKeeper may be a limitation to JobRunr's global processing rate.
* The page request size is hardcoded to 1000

**Describe the solution you'd like**
Allow the configuration of page request sizes used in JobZooKeeper to check for:
* ScheduledJobs
* OrphanedJobs
* SucceededJobs

**Describe alternatives you've considered**
Allow the configuration of each of the mentioned page request sizes. 
* For example, to increase the global processing rate, the ScheduledJobs page request size can be increased, but the remaining queries should not be increased due to performance issues.

These configurations should have the default value of 1000 so that no breaking change is introduced

**Additional context**
For applications that process a high number of _lightweight_ jobs per second, the hardcoded page request size can be a limitation
If we configure the minimum poll interval of 5 seconds, JobRunr will be able to process (on average) 200 jobs per second
* For example, If we have 1750 to be fired in this interval, at most, only 1000 jobs will be updated from Scheduled to Enqueued
* The remaining 750 jobs will only be updated in the next server's poll, but since these are "old" jobs, only 250 new jobs will be fetched from the next scheduling interval
* This will lead to jobs being delayed on applications that constantly execute a high number of jobs per second

Of course, this is dependent on JobRunr's use case and available infrastructure, and therefore the current behaviour should still be kept.

gilberto-p-matos:
JobRunr master batch size configuration for Scheduled, Orphaned and Succeeded Jobs. Fixes #477

### DIFF
--- a/core/src/main/java/org/jobrunr/server/BackgroundJobServerConfiguration.java
+++ b/core/src/main/java/org/jobrunr/server/BackgroundJobServerConfiguration.java
@@ -10,9 +10,13 @@ import java.time.Duration;
 public class BackgroundJobServerConfiguration {
 
     public static final int DEFAULT_POLL_INTERVAL_IN_SECONDS = 15;
+    public static final int DEFAULT_PAGE_REQUEST_SIZE = 1000;
     public static final Duration DEFAULT_DELETE_SUCCEEDED_JOBS_DURATION = Duration.ofHours(36);
     public static final Duration DEFAULT_PERMANENTLY_DELETE_JOBS_DURATION = Duration.ofHours(72);
 
+    int scheduledJobRequestSize = DEFAULT_PAGE_REQUEST_SIZE;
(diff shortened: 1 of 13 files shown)


rdehuyss:
Hi @gilberto-p-matos , if it's ok, I will use your PR to test the newly [contributor license agreement](https://github.com/jobrunr/jobrunr/blob/master/CLA.md) I just added.

Normally, this will only take a minute from your side. To start the CLA process, I must comment on a PR :-).

rdehuyss:
@cla-bot check

PR #493, opened 2022-06-09
Add max back-off interval to RetryFilter

FGummesson:
Hi!
I have a draft of a maximum backoff to the RetryFilter. It sets a roof of the exponential backoff when jobs are scheduled for retries.

Before I continue to add properties to all frameworks, I want to ask. Is this something that you'd consider merging?

### DIFF
--- a/core/src/main/java/org/jobrunr/jobs/filters/RetryFilter.java
+++ b/core/src/main/java/org/jobrunr/jobs/filters/RetryFilter.java
@@ -26,21 +26,28 @@ public class RetryFilter implements ElectStateFilter {
 
     public static final int DEFAULT_BACKOFF_POLICY_TIME_SEED = 3;
     public static final int DEFAULT_NBR_OF_RETRIES = 10;
+    public static final long DEFAULT_MAX_BACKOFF_SECONDS = 0;
 
     private final int numberOfRetries;
     private final int backOffPolicyTimeSeed;
+    private final long maxBackoffSeconds;
 
     public RetryFilter() {
         this(DEFAULT_NBR_OF_RETRIES);
     }
 
     public RetryFilter(int numberOfRetries) {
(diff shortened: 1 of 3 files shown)


rdehuyss:
Hi @FGummesson ! Thanks for the PR! Can I know your functional reason to add this feature?

rdehuyss:
@clabot check

patkujawa-wf:
I'd be interested in this as well. The default of ten retries can cause jobs to run/fail over multiple days, which in our case is undesirable.

patkujawa-wf:
Well, most of our services will heal from an outage fairly quickly and we want jobs to finish as soon as possible, so we'd like to retry until success without having to wait hours between retries. Does that make sense?

rdehuyss:
Well @patkujawa-wf , since you're on Pro, you know you can plugin any retry policy you want? Currently, there is the ExponentialBackoffPolicy and the CustomBackoffPolicy where you can specify the time between each retry. Have you tried that one already?

Perhaps a FixedBackoffPolicy also makes sense?

rdehuyss:
I'm also still interested in the use case of @FGummesson

FGummesson:
We are fetching data from a service using a recurring job. When there is data, jobs runs successfully. When there is no data, I'd like to make use of the back-off. I know that if there is no data at this moment, it will most likely be a while until there is new data to fetch (maybe a few hours).

The recurring job has to run quite often, I've set it to run every 10 seconds currently. 

If i set an exponential back-off with seed 2, and retry 7 times, I will get a maximum backoff of about 2minutes. But on the 7th try I would get an error logged saying the job failed. 10 seconds later, a new job is scheduled and will fail 7 times with a backoff  of 2, 4, 8, 16, 32, 64, 128. 

If something like this `maxBackoff` was added, I wouldn't have to do the first 6 retries to get to and still say that we will get the data within maximum of e.g. 2 minutes.

Issue #525, opened 2022-07-26
[Feature] Make jobrunr JPMS friendly


duponter:
**Is your feature request related to a problem? Please describe.**
I am working on a Spring Boot 2.7 Java 17 application. 

We are planning a feature where we chose jobrunr as part of the solution. Currently, we are using the jobrunr _core_ and the _spring-boot-starter_ modules.

However, as we are JPMS compliant, there are issues when starting our application using the module path.
As there is absolutely no JPMS related configuration present in jobrunr, our application considers the jobrunr jars as a filename-based automodules and automatically exposes all packages from those jars.

Usually this poses no problem, unless multiple jars expose the same package. This is called 'split packages'.
In case of jobrunr, the _core_, the _jobrunr-spring-boot-starter_ and the _jobrunr_quarkus_extension_ modules all expose the _org.jobrunr.scheduling_ package. This implies that these modules cannot be used together on the module path.

**Describe the solution you'd like**
The ideal solution is the introduction of _module-info.java_ in each module. This way JPMS is fully supported and jobrunr can be used both on the module path and the classpath. This also clearly states which packages are exported and therefore are part of the library's api and which packages are considered internal.

This however implies the following changes:
* The source compatibility must be raised to Java 11 (instead of 8).
* Multi Release Jars or an alternative must be provided to keep compatibility with Java 8: https://blog.gradle.org/mrjars
* Each module has its own unique packages to export. Therefore, following classes must be moved to another package:
  - org.jobrunr.scheduling.RecurringJobPostProcessor (in _jobrunr-spring-boot-starter_)
  - org.jobrunr.scheduling.JobRunrRecurringJobRecorder (in _jobrunr_quarkus_extension_) 
  - ... (I might forget some) 

**Describe alternatives you've considered**
An alternative solution is to be aware of JPMS, but not to support it directly. This implies the following actions:
* Each module should provide a manifest file with at least an *Automatic-Module-Name: < chosen module name >* entry to the module's MANIFEST.MF. (http://branchandbound.net/blog/java/2017/12/automatic-module-name/)
* The classes mentioned above must be moved to a unique package as explained earlier.

**Additional context**
I am willing to provide or to contribute to a PR for the JPMS specific configuration and code, but I am not that familiar with the code to make suggestions about moving packages or to decide which packages to expose and which to keep internal.

rdehuyss:
[...]

PR #527, opened 2022-07-29
Added 'Automatic-Module-Name' to the generated MANIFEST.MF

duponter:
Partially implements #525 

For the framework modules, I had to add a **jar** task to the Gradle build files as they didn't exist yet. As I am not that familiar with Gradle and the project, it might be good practice to check configuration I've added.

Also, I did not add that manifest.mf modifications to the **kotlin-1x-support** modules. Please let me know if it has to be added as well to those modules.

Finally, check whether the proposed module names are chosen correctly.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -98,6 +98,7 @@ compileTestFixturesJava {
 jar {
     exclude('**/dependency-check-bin/', '**/dependency-check-reports/')
     manifest {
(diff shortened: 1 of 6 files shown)


rdehuyss:
Hi @duponter , can you configure your git client properly and push again?

duponter:
Replaced by PR #528

PR #543, opened 2022-08-23
fix : Spring AutoConfiguration > jobRunR try to declare metrics even …

w3blogfr:
During unit test, I want to disable jobRunR,

```
org.jobrunr:
 job-scheduler:
  enabled: false
 background-job-server:
  enabled: false
 dashboard:
  enabled: false
```

The problem is JobRunrMetricsAutoConfiguration still try to create StorageProviderMetricsBinder even if there is no StorageProvider and no BackgroundJobServer.

I add a @ConditionalOnBean to avoid that issue.


rdehuyss:
I'm afraid I can't accept this solution as some people just run the metrics on a server that only runs the dashboard.

But, I'm working on another solution for this.

PR #653, opened 2022-12-26
Sql: Fixing statement cache when using tablePrefix

raulgpdev:
The current statementCache does not consider the tablePrefix.

The cache is a static concurrent hashmap, meaning that it will be shared among the different JobScheduler instances.When you configure multiple schedulers with different table prefixes, only one statement is created and cached because the cache key is built using the original query, which does not yet have the table prefix.

As a result, only one tablePrefix is used at the end, while the others are left unused.

Given that the map is static, my suggestion is to simply add the tablePrefix to the map's key.

### DIFF
--- a/core/src/main/java/org/jobrunr/storage/sql/common/db/Sql.java
+++ b/core/src/main/java/org/jobrunr/storage/sql/common/db/Sql.java
@@ -239,7 +239,7 @@ public class Sql<T> {
     }
 
     final String parse(String query) {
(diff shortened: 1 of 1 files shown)


rdehuyss:
Hi @raulgpdev - thanks for this PR but I also need unit tests for this. I don't know yet when I will be able to add them but if you could do it, it will help to get this merged faster.

rdehuyss:
Merged into v6.

Issue #499, opened 2022-06-27
[Feature] Elasticsearch: update to new Java API


jloisel:
Elasticsearch Rest High Level Client is deprecated in favor of:
https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/7.17/index.html

There is still the possibility to have both clients inside a single application:
https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/7.17/migrate-hlrc.html

Still, would be better to migrate to the new client which is the standard for Elasticsearch 8.x.x as well.

jloisel:
Thanks for the feedback. As a workaround, we did:

- Add older sdk to classpath:

```xml
    <dependency>
      <groupId>org.elasticsearch.client</groupId>
      <artifactId>elasticsearch-rest-high-level-client</artifactId>
      <version>${elasticsearch.version}</version>
    </dependency>
```
- Create a Spring configuration to load the `RestHighLevelClient`:

```java
@Configuration
@ConditionalOnProperty(prefix = "org.jobrunr.job-scheduler", name = "enabled", havingValue = "true", matchIfMissing = true)
class JobRunrConfig {

  @Bean
  @SuppressWarnings("deprecated")
  RestHighLevelClient restHighLevelClient(final RestClient client) {
    return new RestHighLevelClientBuilder(client).build();
  }
}
```

jloisel:
We currently use Quartz along with a postgres database. We already have Elasticsearch to store the majority of our data. By switching from Quartz to JobRunr, we can get rid of quartz and the postgresql database.

rdehuyss:
I must confess that Elastic currently is not used by many installations in production - Most people use Postgres.

rdehuyss:
Solved in branch 6.4.

PR #661, opened 2023-01-12
Bug #499: migrate to Elasticsearch new Java API

jloisel:
Hi,

I've spent a few days rewriting the Elasticsearch Storage Provider to use the new Elasticsearch client (as RestHighLevelClient is removed from Elasticsearch 8.x, see https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-high.html).

Other considerations:
- I had to switch to at least java 11 for source and target compatibility otherwise the project wouldn't build on my computer at all (have only Java 17 installed), see build.gradle:

```scala
    sourceCompatibility = JavaVersion.VERSION_11
    targetCompatibility = JavaVersion.VERSION_11
```

So far, i'm facing issues to make junit tests run correctly without any error. I'm trying to run `ElasticSearchJacksonE2ETest`, which fails with Jackson being unable to deserialize the `Job` from json:

```
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `org.jobrunr.jobs.Job` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ...
```

Let me know how we can proceed to integrate those changes into your codebase :)

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,24 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <JetCodeStyleSettings>
+      <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
+    </JetCodeStyleSettings>
+    <codeStyleSettings language="JAVA">
+      <indentOptions>
+        <option name="INDENT_SIZE" value="2" />
+        <option name="CONTINUATION_INDENT_SIZE" value="2" />
+        <option name="TAB_SIZE" value="2" />
+      </indentOptions>
+    </codeStyleSettings>
+    <codeStyleSettings language="XML">
+      <indentOptions>
+        <option name="INDENT_SIZE" value="2" />
+        <option name="CONTINUATION_INDENT_SIZE" value="2" />
+        <option name="TAB_SIZE" value="2" />
+      </indentOptions>
+    </codeStyleSettings>
+    <codeStyleSettings language="kotlin">
+      <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
+    </codeStyleSettings>
+  </code_scheme>
+</component>
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
+    <option name="USE_PER_PROJECT_SETTINGS" value="true" />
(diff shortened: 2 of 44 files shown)


rdehuyss:
Hi @jloisel - before that you put a lot more time in this: why the upgrade to Java 11? Although I know that Java 8 is really old and EOL for active support, currently, too many JobRunr users are still on Java 8.

Does the new ElasticSearch API also support Java 8?

jloisel:
Simply because I have no java 8 on my computer, otherwise the project refuses to build at all. You can switch it back to Java 8 if you need to. You should consider upgrading as there are many interecting new features in Java 11 / 17. That being said, if your customers stay in Java 8 forever, you are a bit stuck.

FIY, we use Java 17 in production on our side.

rdehuyss:
@jloisel - I completely agree with you, I also try to make sure that all my dependencies are up to date (that's why I added the notification in the dashboard when a new JobRunr version is out). 

But oh boy, did my life change when I created this library 😄 .

jloisel:
I hope It changed in the good direction! Same with OctoPerf, it's a life changing experience to create your own software and have lots of people enjoying using it :)

rdehuyss:
I've looked into this but I'm having version conflicts related to jakarta.json-api.

A transitive dependency of `elasticsearch-java` is `jakarta.json-api` which used to be 1.0.11 but now is forced to version `2.0.1`. This causes problems due to the whole namespace change from `javax` to `jakarta` as the `JsonbJsonMapper` (used by Quarkus) uses the older version `1.0.11` with package name starting with `javax`.
Upgrading the `jakarta.json-api` to `2.0.1` is needed for `elasticsearch-java` but thus not compatible with Quarkus 2.x.

Quarkus 3 is slated to be released in the first quarter of 2023 - then we can have a look to upgrade it. Until then, I do not have an idea on how to support both Quarkus and `elasticsearch-java` without a huge refactoring.

image

jloisel:
Can't both can cohabit as the package names are different? It would be an issue if the package names would be clashing. I'm not sure to grasp the issue here, I don't have the extended knowledge you have on your codebase.

I had to add jarkarta json-api 1.1.6 in core/build.gradle otherwise the project wouldn't build at all on my computer:

```
api 'org.glassfish:jakarta.json:1.1.6'
```

Certainly because Elasticsearch was carrying the transitive dependency to your code when using the old java client.

jloisel:
It seems like quarkus is only able to connect to Elasticsearch via the old client :(

https://quarkus.io/guides/elasticsearch

I had to disable the quarkus intellij plugin on my computer as it causes my intellij to regularly freeze, rendering code edition unusable (despite having a Ryzen 5950x + 32GB RAM).

immy (approved):
lgtm

rdehuyss:
Superseded by PR https://github.com/jobrunr/jobrunr/pull/850

PR #793, opened 2023-06-22
feat: update mysql migration v014 script

DatNTHE140692:
Hi, I have a trouble when using jobrunr with other mysql distribution - TiDB, so i create this Pull Request to update migration script to ensure work right.

In this case WITH ROLLUP is not working with TiDB.

Please review for me.

Thanks and Best Regards


rdehuyss:
Hi @DatNTHE140692 ,
- first of all, I do not like it when people push me to review something, especially already after 24h. I'm currently OOO and the good time for a reminder would be after a week.
- you did not consult with me (in a Github discussion or any other way) whether it made sense to add support for TiDB. This would have been nice. At this point, I don't know whether I'm interested in maintaining yet another DB.
- there are not tests for your DB
- then there is already all the feedback from the CLA bot you did not fix yet.
- your changes may result in performance impact for MySQL which I would need to test again but do not have the time for for the moment.

So, please try again and let's see then.

kumarsiva07:
Hi, Any future plan to support TiDB?

Issue #795, opened 2023-06-26
[BUG] Requeue and delete from dashboard broken


geirsagberg:
### JobRunr Version

6.2.2

### JDK Version

OpenJDK 20.0.1

### Your SQL / NoSQL database

Oracle 19c

### What happened?

When clicking either "Requeue" or "Delete" from dashboard, there is an error, and nothing happens.

This seems to be because of this commit: https://github.com/jobrunr/jobrunr/commit/2fb59f1187611949d86d2be2be35780ef685676a

because the `requeue` and `delete` methods still refer to `prop.match.params.id`, not `jobId`. This causes the calls to use `undefined` as jobId.

### How to reproduce?

Go to dashboard, find a failed job, and click Requeue or Delete

geirsagberg:
Workaround: Rolled back to 6.2.1

rdehuyss:
Released with v6.2.3!

Thanks for your PR and please star the repo if you're enjoying JobRunr!

PR #796, opened 2023-06-26
Fix broken requeue and delete

geirsagberg:
fixes https://github.com/jobrunr/jobrunr/issues/795


rdehuyss:
Thanks for your contribution, @geirsagberg! This is the first time I've got such an issue in the frontend... I'm doubting if I should add an extra E2E test to prevent this in the future. 🤔 

@cla-bot check

geirsagberg:
Just use TypeScript and this would be avoided 😉

PR #817, opened 2023-08-17
fix kotlin jackson JobParamererDeserialization

fkowal:
enable configuring the internal ObjectMapper

(with registered KotlinModule) relates to #818 818

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,2 @@
+java openjdk-17
+nodejs 18.12.1
--- a/core/src/main/java/org/jobrunr/utils/mapper/jackson/modules/JobParameterDeserializer.java
+++ b/core/src/main/java/org/jobrunr/utils/mapper/jackson/modules/JobParameterDeserializer.java
@@ -25,9 +25,9 @@ public class JobParameterDeserializer extends StdDeserializer<JobParameter> {
 
     private final ObjectMapper objectMapper;
 
-    protected JobParameterDeserializer() {
+    protected JobParameterDeserializer(ObjectMapper objectMapper) {
         super(JobParameter.class);
-        this.objectMapper = new ObjectMapper();
+        this.objectMapper = objectMapper == null ? new ObjectMapper() : objectMapper;
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


rdehuyss:
@fkowal - if you want me to review this PR, please add yourself to the contributor license agreement.

fkowal:
@rdehuyss https://github.com/jobrunr/clabot-config/pull/23

rdehuyss:
@cla-bot check

.tool-versions
rdehuyss, line 1: Could you exclude this file from the PR, please?

core/src/main/java/org/jobrunr/utils/mapper/jackson/modules/JobParameterDeserializer.java
```diff
@@ -25,9 +25,9 @@ public class JobParameterDeserializer extends StdDeserializer<JobParameter> {
 
     private final ObjectMapper objectMapper;
 
-    protected JobParameterDeserializer() {
+    protected JobParameterDeserializer(ObjectMapper objectMapper) {
         super(JobParameter.class);
-        this.objectMapper = new ObjectMapper();
+        this.objectMapper = objectMapper == null ? new ObjectMapper() : objectMapper;
```
rdehuyss, line 30: I don't see a way that `objectMapper` can be null so I think we can just assign it and the ternary function can be removed.

rdehuyss (commented):
Hi @fkowal, 

By default, JobRunr uses the existing `ObjectMapper` (by means of `jsonParser.getCodec().treeToValue(objectJsonNode, valueType)`) and the `objectMapper` you are changing will only kick in in case of special cases, mainly when an @JsonCreator annotation is used.

It's intentionally a simple `ObjectMapper` to remove all changes done to the JobMapper by JobRunr.

Could you perhaps create a test case that reproduces the issue you are facing (as also I require tests for all changes to JobRunr)?

Thanks!

rdehuyss:
This PR is not valid as there is no issue.

PR #819, opened 2023-08-17
failfast when deserialization was unsuccessful

fkowal:
i can't really tell what the reason to returning a 
```
new JobParameterNotDeserializableException(getActualClassName(className, actualClassName), e.getMessage())
```

this leads to making the jobstate unsound
and unexpected class cast problem in random parts of the codebase (after an async boundary in async pubsub subscription)

### DIFF
--- a/core/src/main/java/org/jobrunr/utils/mapper/jackson/modules/JobParameterDeserializer.java
+++ b/core/src/main/java/org/jobrunr/utils/mapper/jackson/modules/JobParameterDeserializer.java
@@ -44,33 +44,29 @@ public class JobParameterDeserializer extends StdDeserializer<JobParameter> {
     }
 
     private JobParameter getJobParameter(JsonParser jsonParser, String className, String actualClassName, JsonNode objectJsonNode) throws JsonProcessingException {
-        try {
-            Class<Object> valueType = toClass(getActualClassName(className, actualClassName));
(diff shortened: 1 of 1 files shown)


rdehuyss:
This causes the `JobMapperTest.onJobWithParameterThatCannotBeDeserializedAnymoreNoExceptionIsThrown()`. The result is that the dashboard doesn't work anymore as the job cannot be deserialized.

There is definitely a reason that all this code is there 😄 .

Feel free to start a discussion first, I'll give you input why this is done.

fkowal:
still there a case `jobrunr_jobs.jobasjson` state soundness

1. ENQUEUED with sound data = `{...}`
2. PROCESSING deserialization fails replaces the sound data with `JobParameterNotDeserializableException(erromsg)`
so data is lost and there is now way retry the job after a fix is deployed

`class JobParameterNotDeserializedException(JsonNode, error msg)` would at least give you the option to do something about the deserialization error

also the ENQUEUED -> PROCESSING implementation could use the `jsonNode` to not break the job payload

fkowal:
it's hard for me to address the dashboard relation

but `JsonNode` would be very flexible to display job payload in the list / job view
and user / job execution usecase is a very much different

and it worth reconcidering the coupling between these 2 usecases

PR #843, opened 2023-09-25
[Quarkus] Fix metrics producer being added unconditionally

tms0:
This PR fixes an issue where `JobRunrMetricsProducer.StorageProviderMetricsProducer` was always added to the application beans. As a consequence, it was impossible to use the quarkus extension without the `quarkus-micrometer` extension.

### DIFF
--- a/framework-support/jobrunr-quarkus-extension/deployment/src/main/java/org/jobrunr/quarkus/extension/deployment/JobRunrExtensionProcessor.java
+++ b/framework-support/jobrunr-quarkus-extension/deployment/src/main/java/org/jobrunr/quarkus/extension/deployment/JobRunrExtensionProcessor.java
@@ -84,7 +84,6 @@ class JobRunrExtensionProcessor {
         additionalBeans.add(JobRunrProducer.class);
         additionalBeans.add(JobRunrStarter.class);
         additionalBeans.add(jsonMapper(capabilities));
-        additionalBeans.add(JobRunrMetricsProducer.StorageProviderMetricsProducer.class);
         additionalBeans.addAll(storageProvider(capabilities, jobRunrConfiguration));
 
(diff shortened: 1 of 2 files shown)


tms0:
@rdehuyss Another PR for you :) And I will create another small one this week.

rdehuyss:
This just removes the existing bean. Is there a way to find out that the MicroMeter extension is added? I quickly tried to search for it but both MicroMeter and SmallRye provide the `io.quarkus.metrics` capability. See the [Quarkus MicroMeter Extension](https://github.com/quarkusio/quarkus/blob/main/extensions/micrometer/runtime/pom.xml#L149).

As is, I'm afraid I cannot include it as it is a breaking change. If a check and test is added that this is enabled only if MicroMeter is on the classpath, I'm happy to include it.

FYI - I'd like to release soon so if you can update the PR, I'm happy to include it in the release.

rdehuyss (changes requested):
See earlier comment.

tms0:
@rdehuyss  If I understand correctly, the bean is already included by the addMetrics buid step, with all the necessary conditions (https://github.com/jobrunr/jobrunr/blob/master/framework-support/jobrunr-quarkus-extension/deployment/src/main/java/org/jobrunr/quarkus/extension/deployment/JobRunrExtensionProcessor.java#L97), and you had already covered all the cases (https://github.com/jobrunr/jobrunr/blob/master/framework-support/jobrunr-quarkus-extension/deployment/src/test/java/org/jobrunr/quarkus/extension/deployment/JobRunrExtensionProcessorTest.java#L192).

I tried it in a small project, with and without the micrometer extension, it works as expected.

rdehuyss:
You're correct, my mistake. This code was written a long time ago 😊.

PR #845, opened 2023-10-03
[Quarkus] Separate build time and runtime configuration

tms0:
Currently, the configuration is fixed at build time and read-only at runtime (https://quarkus.io/guides/writing-extensions#config-phases). I didn't see it before, but it makes the extension almost unusable when building a native executable. For example, it is not possible to change the datasource credentials at runtime.

This PR separates the build time configuration (i.e. enabling features) and the runtime configuration (i.e. configuring features).


rdehuyss:
Thanks for this - I hope to review it by the end of the week.

A quick question already: doesn't it make sense to have the `MetricsConfiguration` and the `DashboardConfiguration.enabled` in the `JobRunrRuntimeConfiguration`? This way users can enable / disable the dashboard and Metrics at runtime?

In fact, why not do everything in the `JobRunrRuntimeConfiguration`? Isn't the only downside an native executable which is a bit larger?

tms0:
I think this is not the Quarkus philosophy, you want to build exactly what you will run.
An other downside is the build time of the native executable, because each classes will be analyzed, initialized, etc.

In fact, it is already impossible to enable the dashboard at runtime, because the resources are embedded conditionally at build time.

tms0:
But currently, all the dashboard classes are embedded because all the beans are declared in one producer (`JobRunrProducer`). We can optimize it by putting the dashboard things in a separare producer (maybe another PR if you agree with this approch).

tms0:
Hello @rdehuyss ,

This is just a gentle reminder :-)

rdehuyss:
Hi @tms0 - this will be part of JobRunr 6.4 and has an undetermined release date as of now 😄 .

tms0:
Hi @rdehuyss ,

Do you have more visibility about the release date of the 6.4 version ? I don't want to rush you but we really need this change on my project otherwise we'll have to use another library :(

rdehuyss:
Hi @tms0 , I think that we will skip JobRunr 6.4 and move to JobRun 7.0 which will be released early next year.

But I don't understand why you would need to move to a new library - what is keeping you from running your branch temporarily in production?

tms0:
> But I don't understand why you would need to move to a new library - what is keeping you from running your branch temporarily in production?

Hmm, if this is just for a few months, it can be a possiblity.

rdehuyss (approved):
LGTM

rdehuyss:
This is now available in JobRunr 6.3.3.

tms0:
> This is now available in JobRunr 6.3.3.

Thank you so much!

rdehuyss:
Hi @tms0 - we've received the request from multiple users to enable features at runtime (e.g. dashboard, background job server, ...) instead of build time.

I do agree that the build may take a little bit longer but I must confess that I also prefer build once vs having different builds to enable / disable features. 

We're considering to merge PR https://github.com/jobrunr/jobrunr/pull/1093 but we would of course love your input on this also.

tms0:
> Hi @tms0 - we've received the request from multiple users to enable features at runtime (e.g. dashboard, background job server, ...) instead of build time.
> 
> I do agree that the build may take a little bit longer but I must confess that I also prefer build once vs having different builds to enable / disable features.
> 
> We're considering to merge PR #1093 but we would of course love your input on this also.

Hi @rdehuyss ,

I'm really surprised about this request, because doing a maximum of things at build time is one the main idea of Quarkus. Let me quote their website (https://quarkus.io/container-first/) :
> The central idea behind Quarkus is to do at build-time what traditional frameworks do at runtime: configuration parsing, classpath scanning, feature toggle based on classloading, and so on.

> As much processing as possible is done at build time; thus, your application only contains the classes used at runtime. In traditional frameworks, all the classes required to perform the initial application deployment hang around for the application’s life, even though they are only used once. With Quarkus, they are not even loaded into the production JVM! Quarkus does not stop here. During the build-time processing, it prepares the initialization of all components used by your application. It results in less memory usage and faster startup time as all metadata processing has already been done.

In every Quarkus extensions, all these properties are fixed at build time. For example : https://quarkus.io/guides/mongodb#configuration-reference

Ultimately, there is a way to change build time properties at runtime : https://quarkus.io/guides/reaugmentation

rdehuyss:
hi @tms0 - I've opened a discussion on the Quarkus Zulip chat. They don't seem to be against runtime configuration on first sight. Would you be willing to give your [input there](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/JobRunr.20Quarkus.20Extension.20-.20build.20time.20vs.20run.20time.20config)? 

If we change this, I want to make sure we've heard everybody's opinion and do this in line with the Quarkus principles and support from them.

PR #883, opened 2023-11-26
make database migrations work on jboss/wildfly Fixes #383

coschtl:
As Jboss/Wildfly creates URLs with scheme vfs:// when accessing the database migration scripts stored in a jar inside WEB-INF, I added a FilesystemProvider which can handle this.
The files get unpacked by the ClassRessourceProvider and deleted by the FilesystemProvider after usage.


rdehuyss:
@cla-bot check.

core/src/main/java/org/jobrunr/utils/resources/ClassPathResourceProvider.java
```diff
@@ -117,9 +140,10 @@ public void close() throws IllegalStateException {
 
     private FileSystemProvider getFileSystemProviderByScheme(String scheme) {
         switch (scheme) {
-            case "jar": return new JarFileSystemProvider();
-            case "resource": return new ResourcesFileSystemProvider();
-            case "file": return new PathFileSystemProvider();
+            case "jar": return new org.jobrunr.utils.resources.JarFileSystemProvider();
+            case "resource": return new org.jobrunr.utils.resources.ResourcesFileSystemProvider();
+            case "file": return new org.jobrunr.utils.resources.PathFileSystemProvider();
+            case "vfs": return new org.jobrunr.utils.resources.VfsFilesystemProvider();
```
rdehuyss, line 146: Can you import these Provider so they are not fully qualified?

core/src/main/java/org/jobrunr/utils/resources/ClassPathResourceProvider.java
```diff
@@ -77,6 +80,26 @@ private Path toPath(URL url) {
...
+                    Method getPhysicalFileMethod = virtualFileClass.getMethod("getPhysicalFile");
+
+                    List virtualFiles = (List) getChildrenRecursivelyMethod.invoke(virtualFile);
+                    for (Object child : virtualFiles) {
+                        getPhysicalFileMethod.invoke(child);// side effect: create real-world files
+                    }
+                    File rootDir = (File) getPhysicalFileMethod.invoke(virtualFile);
+                    uri = URI.create(org.jobrunr.utils.resources.VfsFilesystemProvider.SCHEME + rootDir.toURI());
+                } catch (InvocationTargetException | NoSuchMethodException | IllegalArgumentException |
+                         IllegalAccessException e) {
```
rdehuyss, line 99: I think we can use `ReflectiveOperationException` here

core/src/main/java/org/jobrunr/utils/resources/ClassPathResourceProvider.java
```diff
@@ -77,6 +80,26 @@ private Path toPath(URL url) {
             if ("wsjar".equals(uri.getScheme())) { // support for openliberty
                 uri = new URI(uri.toString().replace("wsjar", "jar"));
             }
+            if ("vfs".equals(uri.getScheme())) {    // support for Jboss/Wildfly
```
rdehuyss, line 83: Can you use if / else if?

core/src/main/java/org/jobrunr/utils/resources/ClassPathResourceProvider.java
rdehuyss, line 99: Can we extract everything related to VFS to a separate method?

rdehuyss (changes requested):
Some small changes requested... 

Is there a way you think we could test this? I'm afraid this may brake in the future. What do you think?

core/src/main/java/org/jobrunr/utils/resources/ClassPathResourceProvider.java
coschtl: fixed

core/src/main/java/org/jobrunr/utils/resources/ClassPathResourceProvider.java
coschtl: yes, we can if we leave the IllegalArgumentException also ;-)
When creating the fix, I couldn't get my IDE to compile. Therefore I hat to make the changes "blind"

core/src/main/java/org/jobrunr/utils/resources/ClassPathResourceProvider.java
coschtl: sure, if you prefer ;-)

core/src/main/java/org/jobrunr/utils/resources/ClassPathResourceProvider.java
coschtl: done

coschtl:
Regarding testing:
I also thought about this but I think this is not a trivial task.
Only testing the few reflection lines with a mocked virtual file doesn't make much sense as this can not guarantee that a future Wildfly will return virtual file matching our mock.
If we want to assure the Wildfly fix works "forever" we would have to run tests inside an actual Wildfly (i.e. running the test with arquillian). But this leads us to the next problems:
- how to assure that the test always uses the latest Wildfly?
- how to run arquillian tests in gradle (I have absolutely no npm and gradle know how, I even did not get IntelliJ on Windows to fully build the project)?

rdehuyss (approved):
LGTM. I'll merge this in JobRunr v7 which is due Q1 2024 (no exact release date yet).

PR #904, opened 2023-12-27
Initial support for select for update

rdehuyss:


core/src/main/java/org/jobrunr/storage/sql/mariadb/MariaDbDialect.java
```diff
@@ -0,0 +1,32 @@
+package org.jobrunr.storage.sql.mariadb;
+
+import org.jobrunr.storage.sql.common.db.Dialect;
+import org.jobrunr.utils.VersionNumber;
+
+public class MariaDbDialect implements Dialect {
```
auloin, line 6: Is it possible to extend `AnsiDialect` as to avoid some code repetitions? (comment is relevant for most of all newly created `Dialect` implementations)

auloin (approved):

PR #938, opened 2024-02-15
Feature/more cleanup

rdehuyss:


core/src/main/java/org/jobrunr/server/tasks/CheckForNewJobRunrVersion.java
```diff
@@ -40,19 +39,19 @@ public void run() {
...
         } else {
             try {
-                VersionNumber latestVersion = new VersionNumber(getLatestVersion());
-                VersionNumber actualVersion = new VersionNumber(getActualVersion());
+                VersionNumber latestVersion = v(getLatestVersion());
+                VersionNumber actualVersion = v(getActualVersion());
                 if (latestVersion.compareTo(actualVersion) > 0) {
                     dashboardNotificationManager.notify(new NewJobRunrVersionNotification(latestVersion.getCompleteVersion()));
-                    LOGGER.info("JobRunr version {} is available.", latestVersion.completeVersion);
+                    LOGGER.info("JobRunr Pro version {} is available.", latestVersion);
```
auloin, line 54: I think this should be reverted, same for the API URL bellow

auloin (changes requested):

auloin (approved):

PR #958, opened 2024-03-13
Improve tables validation during startup

isabek:
We use Jobrunr to schedule and process dozens of jobs. Restart too much time because of the count operation. Also, Kubernetes kept killing the app because of the liveness checks after the restart. So, the app wasn't able to start properly.

### DIFF
--- a/core/src/main/java/org/jobrunr/storage/sql/common/DatabaseCreator.java
+++ b/core/src/main/java/org/jobrunr/storage/sql/common/DatabaseCreator.java
@@ -91,9 +91,10 @@ public class DatabaseCreator {
              final Transaction tran = new Transaction(conn);
              final Statement pSt = conn.createStatement()) {
             for (String table : JOBRUNR_TABLES) {
-                try (ResultSet rs = pSt.executeQuery("select count(*) from " + tablePrefixStatementUpdater.getFQTableName(table))) {
-                    if (rs.next()) {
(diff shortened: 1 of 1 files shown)


rdehuyss:
I'm afraid we cannot except this PR in it's current form as it only works for MySQL / MariaDB.

What's the reason you have that many jobs? I would advise you to put the have jobs deleted faster after succeeding.

isabek:
@rdehuyss, our policy requires the permanent deletion of jobs after 8 hours. As a result, we may accumulate numerous successful jobs due to this policy (sometimes more than 10M). Personally, I find it somewhat unusual to check table existence using the count operation during initialization. I acknowledge that this approach is generally applicable across most database management systems.

Please don't close this PR for now. I will propose an alternative idea when I have the opportunity.

we-sell-bags:
Why is it even being done this way , by querying the actual table & counting?

I think  getTables(). , is common to all 3 databases since it's in the JDBC standard.

surely simply use "getTables()" with the name of the actual table you are looking at, then query the result set is 1?
or have there been instances of  tables being listed in metadata, but then no actual table?

rdehuyss:
Hi @isabek  - I'm closing this PR as to keep the repo clean. There is not much overhead in creating a new PR.

I think @we-sell-bags is on to something, that would indeed be a better way to check this.

PR #967, opened 2024-03-27
feat : use LettuceConnectionFactory to get RedisClient

rajadilipkolli:
Current implementation forces creation of `RedisClient` Bean when user wants to use spring boot autoconfiguration OOTB.

Change derives already created `RedisClient` from `LettuceConnectionFactory` and becomes pure autoconfiguration without any bean declaration


rajadilipkolli:
Hi @rdehuyss , In my local the PR is compiling. Can you please let me know if I am missing something? I have implemented the hack in my [project ](https://github.com/rajadilipkolli/mfscreener/blob/main/src/main/java/com/learning/mfscreener/config/RedisConfig.java#L12-L14)and it is working as expected

rdehuyss:
Please, do not just create the same PR again. The build fails again and your code cannot compile as there are missing dependencies.

On top of that development now is happening against the v7 branch.

rdehuyss:
P.s.: if you want to create PR's for any project, please checkout the code completely, do your changes and then run tests.

Then you would see what the problem is.

Kind regards,
Ronald

PR #995, opened 2024-04-16
Update JobFilters JavaDoc

auloin:


core/src/test/java/org/jobrunr/jobs/filters/AbstractJobFiltersTest.java
```diff
@@ -24,7 +24,7 @@ void ifJobFilterIsTooSlowAMessageIsLogged() {
         jobCreationFilters.logJobFilterTime(myJobFilter, 11000000);
 
         assertThat(logger).hasWarningMessageContaining(
-                "JobFilter of type 'org.jobrunr.jobs.filters.AbstractJobFiltersTest$MyJobFilter' has slow performance (> 10ms) which negatively impacts the overall functioning of JobRunr",
+                "JobFilter of type 'org.jobrunr.jobs.filters.AbstractJobFiltersTest$MyJobFilter' has slow performance of 11ms (a Job Filter should run under 10ms) which negatively impacts the overall functioning of JobRunr",
```
rdehuyss, line 27: I'm scared about the exact number as it may fail the tests on the CI

rdehuyss (approved):
LGTM!

core/src/test/java/org/jobrunr/jobs/filters/AbstractJobFiltersTest.java
auloin: What would be the issue? The duration is fixed by the test, it should always be exact from my understanding.

PR #997, opened 2024-04-18
Fix spring boot native support

rdehuyss:


framework-support/jobrunr-spring-boot-3-starter/src/main/java/org/jobrunr/spring/aot/JobRunrBeanFactoryInitializationAotProcessor.java
```diff
@@ -162,6 +178,18 @@ private static void registerRequiredJobRunrClasses(RuntimeHints hints) {
                 .registerType(org.jobrunr.jobs.annotations.Job.class, allMemberCategories).registerType(Recurring.class, allMemberCategories)
                 // JobRunr Job Dashboard
                 .registerType(JobDashboardLogger.class, allMemberCategories).registerType(JobDashboardLogger.JobDashboardLogLine.class, allMemberCategories).registerType(JobDashboardLogger.JobDashboardLogLines.class, allMemberCategories).registerType(Problem.class, allMemberCategories).registerType(Page.class, allMemberCategories).registerType(BackgroundJobServerStatus.class, allMemberCategories).registerType(JobStats.class, allMemberCategories).registerType(JobStatsExtended.class, allMemberCategories).registerType(AmountRequest.class, allMemberCategories).registerType(OffsetBasedPageRequest.class, allMemberCategories).registerType(RecurringJobUIModel.class, allMemberCategories).registerType(VersionUIModel.class, allMemberCategories).registerType(SseExchange.class, allMemberCategories);
+
+        if (JAVA_VERSION.hasMajorVersionHigherOrEqualTo(21)) {
```
auloin, line 182:
```suggestion
        if (VirtualThreads.isSupported(JAVA_VERSION))
```

Should we also check that the configured thread type is virtual threads?

auloin (changes requested):

framework-support/jobrunr-spring-boot-3-starter/src/main/java/org/jobrunr/spring/aot/JobRunrBeanFactoryInitializationAotProcessor.java
rdehuyss: Let's see whether we need to do so in other places in the future.

PR #1008, opened 2024-04-23
Better BackgroundJobServer initialization in Spring

SardarNL:
We are facing two minor problems with default JobRunr Spring Boot starter:

- `BackgroundJobServer` is always started before bean is created. A custom `BeanPostProcessor` is unable to configure server before it starts. Fix: use `initMethod = "start"` which was explicitly designed to allow post processors customize bean before/after it is initialized.
- RetryFilter is added using hard-coded list. We are using custom RetryFilter that is aware of domain specifics. Fix: obtain `JobFilter` list from container, by default there is just default `RetryFilter`.

No tests are needed because there is already `JobRunrAutoConfigurationTest.backgroundJobServerAutoConfigurationTakesIntoAccountDefaultNumberOfRetries` which tests for properly configured retry filter.

### DIFF
--- a/framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/spring/autoconfigure/JobRunrAutoConfiguration.java
+++ b/framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/spring/autoconfigure/JobRunrAutoConfiguration.java
@@ -6,6 +6,7 @@ import jakarta.json.bind.Jsonb;
 import org.jobrunr.dashboard.JobRunrDashboardWebServer;
 import org.jobrunr.dashboard.JobRunrDashboardWebServerConfiguration;
 import org.jobrunr.jobs.details.JobDetailsGenerator;
+import org.jobrunr.jobs.filters.JobFilter;
 import org.jobrunr.jobs.filters.RetryFilter;
 import org.jobrunr.jobs.mappers.JobMapper;
 import org.jobrunr.scheduling.JobRequestScheduler;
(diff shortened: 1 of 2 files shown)


SardarNL:
Merge request made using wrong identity yumatov.s@tkppensioen.nl (work e-mail) instead of my private (known to github) SardarNL. Will be closed and re-open again with proper commit.

Thank you cla-bot for noticing.

rdehuyss:
Hi @SardarNL, thanks so much for this PR. Howeber, before you go for the next PR, I would like to give some feedback.

- for the first part (initMethod = "start") I'd like to see the problem it solves first. No other problems were reported related to this. Do you have any exception that you can share or can you give an example repo that reproduces the issue you're seeing. Next to that, we require tests for all these kind of changes.
- for the second part I have some bad news however as this feature is covered by [JobRunr Pro](https://www.jobrunr.io/en/documentation/background-methods/dealing-with-exceptions/#how-to-configure-the-amount-of-retries). Although I really find it difficult to deny this part of the PR, I've spent over 3 years developing JobRunr and I want to prevent open-source burnout (remember the [openssh](https://robmensching.com/blog/posts/2024/03/30/a-microcosm-of-the-interactions-in-open-source-projects/) [vulnerability](https://robmensching.com/blog/posts/2024/03/31/what-could-be-done-to-support-open-source-maintainers/)). That's why I'm afraid I cannot accept it... On top of that, in JobRunr Pro this is covered with over 13 unit and integration tests.

Please don't see this as any criticism on your work but I hope you understand that as you, we would like to get paid for the work we're doing...

PR #1009, opened 2024-04-23
Better BackgroundJobServer initialization in Spring

SardarNL:
We are facing two minor problems with default JobRunr Spring Boot starter:

- `BackgroundJobServer` is always started before bean is created. A custom `BeanPostProcessor` is unable to configure server before it starts. Fix: use `initMethod = "start"` which was explicitly designed to allow post processors customize bean before/after it is initialized.
- RetryFilter is added using hard-coded list. We are using custom RetryFilter that is aware of domain specifics. Fix: obtain `JobFilter` list from container, by default there is just default `RetryFilter`.

No tests are needed because there is already `JobRunrAutoConfigurationTest.backgroundJobServerAutoConfigurationTakesIntoAccountDefaultNumberOfRetries` which tests for properly configured retry filter.

### DIFF
--- a/framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/spring/autoconfigure/JobRunrAutoConfiguration.java
+++ b/framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/spring/autoconfigure/JobRunrAutoConfiguration.java
@@ -6,6 +6,7 @@ import jakarta.json.bind.Jsonb;
 import org.jobrunr.dashboard.JobRunrDashboardWebServer;
 import org.jobrunr.dashboard.JobRunrDashboardWebServerConfiguration;
 import org.jobrunr.jobs.details.JobDetailsGenerator;
+import org.jobrunr.jobs.filters.JobFilter;
 import org.jobrunr.jobs.filters.RetryFilter;
 import org.jobrunr.jobs.mappers.JobMapper;
 import org.jobrunr.scheduling.JobRequestScheduler;
@@ -38,6 +39,8 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 
+import java.util.List;
+
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static java.util.Optional.ofNullable;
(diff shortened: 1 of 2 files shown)


rdehuyss:
Hi @SardarNL , I don't know whether you saw my feedback on the closed PR. I'll also put it here:

Hi @SardarNL, thanks so much for this PR. Howeber, before you go for the next PR, I would like to give some feedback.

- for the first part (initMethod = "start") I'd like to see the problem it solves first. No other problems were reported related to this. Do you have any exception that you can share or can you give an example repo that reproduces the issue you're seeing. Next to that, we require tests for all these kind of changes.
- for the second part I have some bad news however as this feature is covered by [JobRunr Pro](https://www.jobrunr.io/en/documentation/background-methods/dealing-with-exceptions/#how-to-configure-the-amount-of-retries). Although I really find it difficult to deny this part of the PR, I've spent over 3 years developing JobRunr and I want to prevent open-source burnout (remember the [openssh](https://robmensching.com/blog/posts/2024/03/30/a-microcosm-of-the-interactions-in-open-source-projects/) [vulnerability](https://robmensching.com/blog/posts/2024/03/31/what-could-be-done-to-support-open-source-maintainers/)). That's why I'm afraid I cannot accept it... On top of that, in JobRunr Pro this is covered with over 13 unit and integration tests.

Please don't see this as any criticism on your work but I hope you understand that as you, we would like to get paid for the work we're doing...

SardarNL:
Hi @rdehuyss thanks for replying.

> for the first part (initMethod = "start") I'd like to see the problem it solves first.

We need programmatic fine tuning of `BackgroundJobServer` and one way to do it is to define `BeanPostProcessor`. Spring defines life-cycle graph for all beans:

1. Create bean
2. Call optional `BeanPostProcessor.postProcessBeforeInitialization()`
3. Initialize bean by calling `afterPropertiesSet()`, `@PostConstruct` and init method.
4. Call optional `BeanPostProcessor.postProcessAfterInitialization()`

At this moment job server gets started in step 1. This is in violation of Spring's life-cycle management and provides minor nuisance for developers.

> for the second part I have some bad news however as this feature is covered by [JobRunr Pro](https://www.jobrunr.io/en/documentation/background-methods/dealing-with-exceptions/#how-to-configure-the-amount-of-retries).

Pro version provides flexible retry using configuration. Here we make it possible for programmatic configuration of job filters - a typical best-practice when containers are used (in this case Spring container). This is not a blokker for anyone, just a minor nuisance. We are able to override `JobRunrAutoConfiguration.backgroundJobServer()` completely to initialize `BackgroundJobServer` with our own filters and not start it when it is needed. So hiding this change behind a paywall isn't really viable.

I'm not the right person to suggest which options should be offered at a price, but I believe it should be about quality of life options, such as better UI, better monitoring, ability to run jobs in the cloud, runtime configuration of workers/master and beter insights in overall performance of the cloud. I don't believe couple of lines of code replacing `BackgroundJobServer` bean with custom one will force anyone into buying Pro version. I believe the opposite will happen because of possible frustration/lack of trust.

rdehuyss:
Already some feedback:
- The `initMethod` was not done consistently (webserver has same issue). I created PR #1012 to fix that already as we want to release soon.
- I need to think about your input regarding the `RetryFilter`. I can understand your frustration but I've been told many times that JobRunr OSS is feature complete and already too good to upgrade to Pro. These small frustrations can help in convincing to upgrade. And I totally understand if you override the `BackgroundJobServer` but I also hope you understand our point of view... I've 3 kids to feed and entertain and I work about 80h a week so the whole OSS developer burnout thing is definitely an issue.

rdehuyss:
Quick update related to:
> Related to RetryFilter, by default there is no way to tell JobRunr not to retry a task if certain exception is raised. There are transient failures (almost all of them) and failures where retry would not help (eg. rejection of a message because it fails validation). We use our own filter to intercept that kind of "critical" exceptions and not retry.

You can just throw a `JobRunrException` where you pass true for the boolean `doNotRetry` e.g.:
```java
new JobRunrException("Job failed because of blabla", true)
```

SardarNL:
You are right, throwing `JobRunrException` will prevent JobRunr from retrying. Unfortunately, it will also vendor lock-in us to JobRunr and I totally understand why you would like it to happen.

Our tasks are scheduled from domain module, which is kept free from infrastructure. I wrote a simple proxy wrapper in tasks module:
```java
    public <T> T of(T bean) {
        var targetClass = AopUtils.getTargetClass(bean);
        var strippedOfAllOtherAdvice = AopProxyUtils.getSingletonTarget(bean);
        var proxy = new ProxyFactory(strippedOfAllOtherAdvice == null ? bean : strippedOfAllOtherAdvice);
        proxy.addAdvice(0, new TaskMethodInterceptor(targetClass));
        return (T) proxy.getProxy();
    }
```
Where `TaskMethodInterceptor` intercepts all methods and schedles them on JobRunr. This way we can use:
```java
tasks.of(anyService).anyMethod(); // this method will be scheduled on JobRunr
```
Without depending on `@Job` and `JobScheduler`. Such method may throw a specific "critical" domain exception which prevents underlying tasks queue from from retrying - that is why I needed custom retry filter.

I totally understand why container configured job filters may provide too much flexibility and why you wouldn't like it to happen. With proper initialization we got at least some of our needs sorted, thanks for that.

rdehuyss:
Well, the general idea is that JobRunr is so light that even with depending only on `JobScheduler`, `@Job` and `@Recurring` and the `JobRunrException`, you could still easily switch to another solution. The last thing I want is for you to have a vendor lock-in.

Thanks for your understanding - I hope you continue to enjoy JobRunr!

PR #1012, opened 2024-04-23
Improve spring integration thanks to feedback from @SardarNL.

rdehuyss:


bmertens911 (approved):
Also approved by @nicholasM95 :-)

PR #1058, opened 2024-06-20
refactoring the code

rohankumardubey:
refactoring the code

### DIFF
--- a/core/src/main/java/org/jobrunr/storage/InMemoryStorageProvider.java
+++ b/core/src/main/java/org/jobrunr/storage/InMemoryStorageProvider.java
@@ -130,7 +130,7 @@ public class InMemoryStorageProvider extends AbstractStorageProvider {
                 .filter(entry -> entry.getValue().getLastHeartbeat().isBefore(heartbeatOlderThan))
                 .map(Map.Entry::getKey)
                 .collect(toList());
-        backgroundJobServers.keySet().removeAll(serversToRemove);
+        serversToRemove.forEach(backgroundJobServers.keySet()::remove);
         return serversToRemove.size();
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jobrunr/storage/InMemoryStorageProvider.java
```diff
@@ -130,7 +130,7 @@ public int removeTimedOutBackgroundJobServers(Instant heartbeatOlderThan) {
                 .filter(entry -> entry.getValue().getLastHeartbeat().isBefore(heartbeatOlderThan))
                 .map(Map.Entry::getKey)
                 .collect(toList());
-        backgroundJobServers.keySet().removeAll(serversToRemove);
+        serversToRemove.forEach(backgroundJobServers.keySet()::remove);
```
rdehuyss, line 133: I don't think it is on us to fix or workaround this (as the JVM/JDK here is slow and it may be fixed in future releases). 
Also, as the `backgroundJobServers` will always be larger than the `serversToRemove` the issue mentioned by Intellij does not impact us.

core/src/main/java/org/jobrunr/storage/InMemoryStorageProvider.java
```diff
@@ -197,7 +197,7 @@ public void deleteMetadata(String key) {
                 .map(JobRunrMetadata::getId)
                 .collect(toList());
         if (!metadataToRemove.isEmpty()) {
-            this.metadata.keySet().removeAll(metadataToRemove);
+            metadataToRemove.forEach(this.metadata.keySet()::remove);
```
rdehuyss, line 200: Same comment as above.

core/src/main/java/org/jobrunr/storage/InMemoryStorageProvider.java
```diff
@@ -233,7 +233,7 @@ public int deleteJobsPermanently(StateName state, Instant updatedBefore) {
                 .filter(job -> job.getUpdatedAt().isBefore(updatedBefore))
                 .map(Job::getId)
                 .collect(toList());
-        jobQueue.keySet().removeAll(jobsToRemove);
+        jobsToRemove.forEach(jobQueue.keySet()::remove);
```
rdehuyss, line 236: Same comment as above.

rdehuyss (changes requested):
Hi @rohankumardubey!

Thanks for this. We mostly agree with Intellij except for your first changes...

Any reason why you opened this?

rdehuyss:
Closing due to no updates / answers

PR #1083, opened 2024-08-12
Set log level to info for recurring job already scheduled

michaelcowan:
I propose we change the following log line from DEBUG to INFO:
* `"Run will be skipped as job is taking longer than given CronExpression or Interval."`

My reasoning here is that, looking at the other log lines, it seems that DEBUG is used for the common (often expected) path e.g., 
* `"Recurring job '{}' resulted in 1 scheduled job."`

The less likely to occur and (in my head at least) more important logs seem to be INFO e.g.,
* `"This means a long GC happened and JobRunr is catching up"`

Id argue that skipping runs belongs in the INFO group of less noisy but more informative logs.


michaelcowan:
I opened a PR to add myself as a contributor https://github.com/jobrunr/clabot-config/pull/42

rdehuyss:
@cla-bot check

rdehuyss (approved):

PR #1087, opened 2024-08-14
Add scm information to quarkus-extension.yaml

### LINKED ISSUE #1085: [Feature] Allow Quarkus tooling to locate the source repository for the extension (ie this repo)
Most extensions on https://quarkus.io/extensions have enriched information (icon, issue counts, etc) which we pull from GitHub. That information is missing for https://quarkus.io/extensions/org.jobrunr/quarkus-jobrunr/ because the registry don't have enough information to work out the source repo. 

The web page gets the repo location from scm coordinates in the enriched copy of the extension metadata file. That information gets into the file by a few mechanisms: 
- if the extension is built using GitHub Actions the build tooling pulls the information from the the `GITHUB_REPOSITORY` environment variables
- otherwise, the build tooling also read it from the `scm` information in the pom.xml
- if the`quarkus-extension.yaml` has the information already, it will be used

It would be great if the JobRunr extension also has this information available, and it's fairly easy to get it there. I haven't done a PR since I'm not sure which mechanism you'd prefer. I'm not 100% sure setting `GITHUB_REPOSITORY` in the environment will work, since this is a gradle extension. Setting an `<scm>` in the `pom.xml`  definitely won't work. :) But setting something in the `quarkus-extension.yaml` would work. 

It would be something like this:

```
metadata:
 [...]
  scm-url: "https://github.com/jobrunr/jobrunr"
```

If you wanted it to be dynamic, resource filtering or equivalent might work. 

To check it's working, you can have a peek in the generated quarkus-extension.yml in the runtime project to make sure there's an scm field.

holly-cummins:
Resolves #1085.

### DIFF
--- a/framework-support/jobrunr-quarkus-extension/quarkus-jobrunr/src/main/resources/META-INF/quarkus-extension.yaml
+++ b/framework-support/jobrunr-quarkus-extension/quarkus-jobrunr/src/main/resources/META-INF/quarkus-extension.yaml
@@ -14,6 +14,7 @@ metadata:
   categories:
     - "miscellaneous"
   status: "stable"
+  scm-url: "https://github.com/jobrunr/jobrunr"
   extension-dependencies:
     - "io.quarkus:quarkus-arc"
     - "io.quarkus:quarkus-core"


rdehuyss:
@cla-bot check

rdehuyss (approved):
Thx!

PR #1104, opened 2024-08-29
Fix log message to avoid confusion

ShadowOfLies:
Avoids the confusing logs as seen here.
image


rdehuyss:
TBH, I wonder whether we even need the first log message? We're logging duplicate info.

Is it not better to remove it completely?

ShadowOfLies:
Well, I'm not as familiar with the inner workings, but what I do see is that the second log has an additional condition before it is logged, but the shutdown wait only applies if the condition passes.

Perhaps changing the first log to be a more general "shutting down" message.  Something simple like `Stopping BackgroundJobServer` or `BackgroundJobServer shutting down`

rdehuyss:
Thx for this! Closing in favor of #1117.

PR #1119, opened 2024-09-25
Improve Quarkus buildtime and runtime properties separation

auloin:


auloin:
Hi @tms0 and @prabhjotlalli, we'd like feedback from the two of you on this approach. We have moved the `enabled` property to runtime and replaced it with a new buildtime property `included` allowing us better control over the included resources. Does this cover your use cases?

rdehuyss (approved):
LGTM!

PR #1140, opened 2024-10-17
Fix concurrent jobState change during save.

rdehuyss:
Try to fix issue since v7 where jobs are changed while being saved to the DB. This results in a different state in the STATE column than the jobAsJson column.


auloin (approved):

rdehuyss:
Fixes https://github.com/jobrunr/jobrunr-pro/issues/455.

Issue #1139, opened 2024-10-16
[BUG] Migration Lock Row Not Removed After Migration


tomdeering-wf:
### JobRunr Version

7.3.1

### JDK Version

Corretto 21

### Your SQL / NoSQL database

Postgres 14

### What happened?

Dependabot recently upgraded our version of `org.jobrunr:jobrunr-pro-spring-boot-3-starter` from 7.2.3 to 7.3.0, and subsequently our service failed to deploy to several environments. Investigating, a prior migration run in July by version v7.2.2 ran a migration but failed to clean up its migration lock row (record with `id = 00000000-0000-0000-0000-000000000000` in `jobrunr_migrations`). Version 7.3.0 was also attempting to run a migration and failed because it could not acquire the migration lock.

![image](https://github.com/user-attachments/assets/78436616-8304-473f-8103-c393e1f930e1)

In order to unblock our service deploying with the new version of jobrunr we had to manually remove the lock row. It seems like there is an opportunity for jobrunr to be a more resilient so this does not occur again. This happened in several of our environments, so it seems likely to be a not-super-rare failure mode.

### How to reproduce?

Deploy service with jobrunr pro 7.2.2 using Postgres 14, observe that the table locker row is not cleaned up. If necessary, interrupt the service while jobrunr is migrating... I'm not sure whether this occurred in our environments.

### Relevant log output

```shell
exception: { [-]
     message: Database migrations have timed out.
     stacktrace: java.lang.IllegalStateException: Database migrations have timed out.
 at org.jobrunr.storage.sql.common.DatabaseCreator$MigrationsTableLocker.isMigrationsTableLocked(DatabaseCreator.java:422)
 at org.jobrunr.storage.sql.common.DatabaseCreator$MigrationsTableLocker.waitUntilMigrationsAreDone(DatabaseCreator.java:400)
 at org.jobrunr.storage.sql.common.DatabaseCreator$MigrationsTableLocker.access$300(DatabaseCreator.java:344)
 at org.jobrunr.storage.sql.common.DatabaseCreator.runMigrations(DatabaseCreator.java:164)
 at org.jobrunr.storage.sql.common.DatabaseCreator.runMigrations(DatabaseCreator.java:120)
 at org.jobrunr.storage.sql.common.DefaultSqlStorageProvider.setUpStorageProvider(DefaultSqlStorageProvider.java:89)
[...]

PR #1142, opened 2024-10-21
Prevent servers from starting if the MigrationsTable is still locked

auloin:


tomdeering-wf:
[Xposting](https://github.com/jobrunr/jobrunr/issues/1139#issuecomment-2427473771) for visibility. Would this change potentially cause unnecessary outages? I think we'd only want to refuse to start if we believe it's unsafe to do so. If we don't have migrations that need to run, is it truly unsafe for jobrunr to start on this server? I don't think so.

rdehuyss:
Hi @tomdeering-wf - we did add an extra check by first seeing if the Migrations table is there by means of querying the database metadata and if so, we do not catch the DB exception anymore.

rdehuyss (approved):
LGTM!

Issue #977, opened 2024-04-01
[Feature] Automatically create jobs if Spring Bean is annotated with @Job annotation


rdehuyss:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I need to write boilerplate code. I'm not a fan of always repeating:

```java
jobScheduler.enqueue(() -> emailClient.send(to));
```

**Describe the solution you'd like**
The idea would be that you could annotate an interface and spring could create a proxy so that when you invoke a method on an implementation of that interface it schedules a job for you (to be run).

```java
@JobGateway // or @Async
interface EmailClient {
 @Job(name =“sendEmail”) 
 void send(String to); 
}
```

**Describe alternatives you've considered**
🤔 can't think of anything else...

**Additional context**
Reported by @joshlong

devashishTaneja:
Hi @rdehuyss, is this feature open for contribution?

rdehuyss:
Yes, of course. Feel free to ping me if you want to discuss ideas on how to implement this.

devashishTaneja:
Hi @rdehuyss,

I have raised an initial PR for this. Please let me know if it looks good.

https://github.com/jobrunr/jobrunr/pull/1181

rdehuyss:
Resolved in V8 branch. Thanks, @devashishTaneja!

PR #1181, opened 2025-01-06
[Feature] Automatically create jobs if Spring Bean is annotated with @AsyncJob Annotation

devashishTaneja:
[Feature] Automatically create jobs if Spring Bean is annotated with @Job annotation #977


rdehuyss:
Hi @devashishTaneja - I noticed you did not add yourself as a contributor yet. Can you do so, that way we can take a look at the PR.

Thx!

devashishTaneja:
Hi @rdehuyss, raised PR to clabot.
https://github.com/jobrunr/clabot-config/pull/51/files

rdehuyss:
@cla-bot check

rdehuyss:
I quickly tested it but it does not work if you run it within a Spring Boot App.

Here is some code that works but needs cleaning up:
```java
package org.jobrunr.scheduling;

import org.jobrunr.jobs.JobDetails;
import org.jobrunr.jobs.JobParameter;
import org.jobrunr.jobs.annotations.Job;
import org.jobrunr.jobs.annotations.JobGateway;
import org.jobrunr.server.runner.ThreadLocalJobContext;
import org.jobrunr.utils.JobUtils;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.beans.BeansException;
import org.springframework.beans.factory.BeanFactory;
import org.springframework.beans.factory.BeanFactoryAware;
import org.springframework.beans.factory.config.BeanPostProcessor;
import org.springframework.cglib.proxy.Enhancer;
import org.springframework.cglib.proxy.MethodInterceptor;
import org.springframework.cglib.proxy.MethodProxy;

import java.lang.reflect.Method;
import java.util.Arrays;
import java.util.List;

import static java.util.stream.Collectors.toList;


public class JobGatewayPostProcessor implements BeanPostProcessor, BeanFactoryAware {

    private static final Logger LOGGER = LoggerFactory.getLogger(JobGatewayPostProcessor.class);
    private BeanFactory beanFactory;

    @Override
    public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
        if (bean.getClass().isAnnotationPresent(JobGateway.class)) {
            Enhancer enhancer = new Enhancer();
            enhancer.setSuperclass(bean.getClass());
            enhancer.setCallback(new MethodInterceptor() {
                @Override
                public Object intercept(Object obj, Method method, Object[] args, MethodProxy proxy) throws Throwable {
                    if (!method.isAnnotationPresent(Job.class)) {
                        // it's not a job so we're not interested
                        return proxy.invoke(bean, args);
                    } else if (ThreadLocalJobContext.getJobContext() != null) {
                        // we're being invoked from within a Job. Do not create a job otherwise we have an infinite loop
                        return proxy.invoke(bean, args);
                    }


                    JobScheduler jobScheduler = beanFactory.getBean(JobScheduler.class);
                    String declaringClassName = method.getDeclaringClass().getName();
                    String methodName = method.getName();
                    List<JobParameter> jobParameters = Arrays.stream(args).map(JobParameter::new).collect(toList());
                    JobDetails jobDetails = new JobDetails(declaringClassName, null, methodName, jobParameters);
                    jobDetails.setCacheable(true);
                    LOGGER.info("Creating job via JobGateway: " + JobUtils.getJobSignature(jobDetails));
                    jobScheduler.enqueue(null, jobDetails);

                    return null;
                    // Execute the original method
                    //return proxy.invoke(bean, args);
                }
            });
            return enhancer.create();
        }
        return bean;
    }

    @Override
    public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
        this.beanFactory = beanFactory;
    }

}
```

Things that still need to be done:
- this `JobGatewayPostProcessor` needs to be a bean and be created by the JobRunrAutoConfiguration (see also the `RecurringJobPostProcessor`) as it otherwise cannot work
- on startup, validation needs to be added: if the method annotated with @Job returns anything (e.g. it is not a void method), we need to throw an exception as otherwise it may cause confusion. We cannot return something if it is being executed by JobRunr in a different thread
- tests for this including an end to end test.

Things to discuss:
- do we like the @JobGateway or do we want to reuse the `@Async` annotation from Spring? Personally, I like the `@Async` but Quarkus does not have this concept. Should we introduce `@AsyncJob`?

rdehuyss:
Note: I just found out cglib is not further being developped...

devashishTaneja:
Hi @rdehuyss, I have the appropriate changes and tested with the below sample application.
https://github.com/devashishTaneja/jobrun-test

rdehuyss:
Hi @devashishTaneja - I think ChatGPT may have rolled me. It said Spring was transitioning to ByteBuddy but when I looked into the code, I don't see any prove of that. To be sure what to use, I asked this [stackoverflow question](https://stackoverflow.com/questions/79373618/spring-and-cglib-future).

If we don't get an answer from the spring team, I'll reach out to some of the core developers directly. Sorry for the confusion.

mdeinum:
Is there a reason to use CGLIB directly. This code seems already tied to Spring anyway so I would suggest to instead use the `ProxyFactory` and use a `MethodInterceptor`. This makes proxy creation transparent and doesn't really care which proxy mechanism is being used.

mdeinum:
I don't see why you would need a proxy for that, as stated in the `SmartInitializingSingleton`. Assuming that all beans have been annotated with `@JobGateway` you can simply look them up (as I'm doing in there), then iterate over the methods (as done as well) and create jobs for the methods annotated with `@AsyncJob`. If that is all you need to do I really don't see the need for the proxies at all. 

This is basically also how `@Scheduled`, for instance, is working. It detects beans with methods having that annotation and then registers them with a `TaskScheduler`. This (from my point of view) is more or less the same.

devashishTaneja:
Hi @rdehuyss, I have switched to ProxyFactory and am able to achieve the goal. Please have a look once.

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/AsyncJobPostProcessor.java
```diff
@@ -0,0 +1,49 @@
...
+            }
+        }
+        return bean;
+    }
+
+    public Object applyJobEnhancement(Object bean) throws InstantiationException, IllegalAccessException, NoSuchMethodException, InvocationTargetException {
+        JobScheduler jobScheduler = beanFactory.getBean(JobScheduler.class);
+        ProxyFactory proxyFactory = new ProxyFactory(bean);
+        proxyFactory.setProxyTargetClass(true);
+        proxyFactory.addAdvice((org.aopalliance.intercept.MethodInterceptor) invocation -> {
```
mdeinum, line 34: Why not just implement `MethodInterceptor` in the `JobInterceptor`? You would only need one single instance of the interceptor anyway. This would apply for the Spring Boot 3 one as well (wouldn't there be a way to share this code instead of duplicating it?).

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/AsyncJobPostProcessor.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.lang.reflect.InvocationTargetException;
+
+@Component
+public class AsyncJobPostProcessor implements BeanPostProcessor, BeanFactoryAware {
+    BeanFactory beanFactory;
+
+    @Override
+    public Object postProcessBeforeInitialization(Object bean, String beanName) {
+        if (bean.getClass().isAnnotationPresent(AsyncJob.class)) {
```
mdeinum, line 20: I would suggest using `AnnotatedElementUtils.hasAnnotation(bean.getClass(), AsyncJob.class)`. That would take into account proxies, annotations on interfaces or even composed annotations if this is ever going to be used as such.

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/AsyncJobPostProcessor.java
```diff
@@ -0,0 +1,49 @@
...
+            try {
+                return applyJobEnhancement(bean);
+            } catch (InstantiationException | IllegalAccessException | NoSuchMethodException | InvocationTargetException e) {
+                throw new RuntimeException(e);
+            }
+        }
+        return bean;
+    }
+
+    public Object applyJobEnhancement(Object bean) throws InstantiationException, IllegalAccessException, NoSuchMethodException, InvocationTargetException {
```
mdeinum, line 30: Are these exceptions thrown from the code inside this method? As I don't see it. So they could be removed as well as the catch from the `postProcessBeforeInitialization`.

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/JobInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+
+        // If running within a job context or method is not annotated with @Job, proceed as usual
+        if (ThreadLocalJobContext.getJobContext() != null || !method.isAnnotationPresent(Job.class)) {
+            return invocation.proceed();
+        }
+
+        // Extract job details
+        String declaringClassName = method.getDeclaringClass().getName();
+        String methodName = method.getName();
+        List<JobParameter> jobParameters = Arrays.stream(args)
```
mdeinum, line 38: Judging from the exception thrown earlier, this list will always be empty.

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/JobInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+        this.jobScheduler = jobScheduler;
+    }
+
+    public Object intercept(org.aopalliance.intercept.MethodInvocation invocation) throws Throwable {
+        Method method = invocation.getMethod();
+        Object[] args = invocation.getArguments();
+
+        // If arguments are present, throw an exception (JobRunr limitation)
+        if (args.length > 0) {
+            throw new RuntimeException("JobRunr does not support method arguments");
```
mdeinum, line 27: Spring prefers more specific exceptions then `RuntimeException`. So something like an `IllegalStateException` or maybe JobRunR has some more specific exception.

mdeinum (commented):
Looks good overall (from a Spring perspective). Just some comments to make it integrate just a bit better.

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/JobInterceptor.java
devashishTaneja: Fixed

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/JobInterceptor.java
devashishTaneja: Makes sense, fixed.

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/AsyncJobPostProcessor.java
devashishTaneja: Fixed

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/AsyncJobPostProcessor.java
devashishTaneja: Fixed

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/AsyncJobPostProcessor.java
devashishTaneja: I fixed the first issue, but I am unsure about the second part about duplicating code followed the same way as it is done in other places.
@rdehuyss your input can help here.

devashishTaneja:
@mdeinum - Thanks for the inputs. Updated the code please have a look.

rdehuyss:
Can I say that I love how things are going here? This is the first time I've seen such a great collaboration of different OSS users on a PR in JobRunr. 🥳 🙇‍♂️ 

I'll also review later today! Thanks again @devashishTaneja  and @mdeinum !

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/JobInterceptor.java
```diff
@@ -0,0 +1,53 @@
...
+    @Nullable
+    @Override
+    public Object invoke(@Nonnull MethodInvocation invocation) throws Throwable {
+        Method method = invocation.getMethod();
+        Object[] args = invocation.getArguments();
+
+        // If arguments are present, throw an exception (JobRunr limitation)
+        if (args.length > 0) {
+            throw new IllegalStateException("JobRunr does not support method arguments");
+        }
```
rdehuyss, line 33: For this type of job (a normal `job.enqueue()`, we do accept method arguments. 

It's only for recurring jobs that method arguments don't make sense. So, we can definitely have arguments.

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/JobInterceptor.java
```diff
@@ -0,0 +1,53 @@
...
+
+        // If running within a job context or method is not annotated with @Job, proceed as usual
+        if (ThreadLocalJobContext.getJobContext() != null || !AnnotatedElementUtils.hasAnnotation(method, Job.class)) {
+            return invocation.proceed();
+        }
+
+        // Extract job details
+        String declaringClassName = method.getDeclaringClass().getName();
+        String methodName = method.getName();
+        JobDetails jobDetails = new JobDetails(declaringClassName, null, methodName, new ArrayList<>());
```
rdehuyss, line 43: To add the parameters, you can just use `Stream.of(args).map(JobParameter::new).collect(Collectors.toList))`

rdehuyss (changes requested):
Looks good! 

I think we can add the paramaters. I also added an integration test in spring boot 3 for it (we have troubles with integration tests in spring boot 2 due to SLF4J changes).

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/JobInterceptor.java
devashishTaneja: I understand this makes sense to me, I wasn't aware of this use case.
@rdehuyss - Just to clarify, do we need to throw the exception only if the method is annotated with @Recurring?

rdehuyss: Yes, indeed. 

But, the @Recurring is handled by the `RecurringJobPostProcessor` and out of scope for the `@AsyncJobPostProcessor`. The exception is already being thrown in the `RecurringJobPostProcessor`.

Does that make sense?

devashishTaneja: Yes, this makes sense. Thanks for the input

devashishTaneja: @rdehuyss - Made the changes to support parameters as well and updated the unit test for the same.

mdeinum:
LGTM. I don't see anything wrong with this from a Spring perspective.

auloin (commented):
@devashishTaneja thanks for the good work! I only have a few comments on the code itself.

More general remarks:
- I'll be retargeting the PR to the `v8` branch, as we plan to release the feature in JobRunr v8. Could you also update your branch with the latest in `v8`?
- Should `@AsyncJob` be on the method and not the class? Or both? What do you think @rdehuyss?

framework-support/jobrunr-spring-boot-2-starter/src/main/java/org/jobrunr/scheduling/AsyncJobPostProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+import org.jobrunr.jobs.annotations.AsyncJob;
+import org.springframework.aop.framework.ProxyFactory;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.beans.factory.BeanFactoryAware;
+import org.springframework.beans.factory.config.BeanPostProcessor;
+import org.springframework.core.annotation.AnnotatedElementUtils;
+import org.springframework.stereotype.Component;
+
+@Component
```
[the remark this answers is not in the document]
mdeinum: If you already have an `@Bean` this annotation can be removed. Saves you also from the risk of getting 2 instances by accident.

devashishTaneja:
@auloin, @rdehuyss - Improved based on the above review comments, and rebased to v8.

mdeinum (approved):

rdehuyss:
Thx for the great work - I did some small improvements and it is ready to be merged.

auloin (approved):

rdehuyss (approved):
LGTM!

PR #1185, opened 2025-01-14
Auto delete RecurringJob's via annotation.

rdehuyss:
When a RecurringJob that is created by an annotation is deleted from the code, it will automatically be removed from the StorageProvider.


core/src/main/java/org/jobrunr/scheduling/ScheduleExpressionType.java
```diff
@@ -30,5 +35,13 @@ public static Schedule getSchedule(String scheduleExpression) {
         throw new ScheduleException(scheduleExpression);
     }
 
+    public static String findSchedule(String cron, String interval) {
```
auloin, line 38: I find that `findSchedule` doesn't describe well what the method does. What do you think of `selectConfiguredScheduleExpression`? Could also become more generic and accept any amount of parameters.

framework-support/jobrunr-micronaut-feature/src/main/java/org/jobrunr/scheduling/JobRunrRecurringJobScheduler.java
```diff
@@ -36,13 +34,9 @@ public void schedule(ExecutableMethod<?, ?> method) {
         String id = getId(method);
         String cron = getCron(method);
         String interval = getInterval(method);
+        String scheduleAsString = ScheduleExpressionType.findSchedule(cron, interval);
```
auloin, line 37: For consistency: I think we already use `scheduleExpression` to refer to the `Schedule` in its string format.

auloin (commented):
Nice job! I added a few comments on some specific areas.

I'd also want your opinion on deleting the `RecurringJob` if the annotation is removed from the method. Is this possible or desirable (performance impact)?

rdehuyss:
Hi @auloin - can you quickly review again? 

If a job is not annotated with the `@Recurring` annotation anymore, it will automatically be deleted.

core/src/test/java/org/jobrunr/utils/JobUtilsTest.java
```diff
@@ -34,4 +34,12 @@ void testGetJobAnnotation() {
         assertThatCode(() -> JobUtils.getJobAnnotation(classThatDoesNotExistJobDetails().build())).doesNotThrowAnyException();
         assertThat(logger).hasWarningMessageContaining("Trying to find Job Annotations for 'i.dont.exist.Class.notImportant(java.lang.Integer)' but the class could not be found. The Job name and other properties like retries and labels will not be set on the Job.");
     }
+
+    @Test
+    void testGetRecurringAnnotation() {
```
auloin, line 39: This test hasn't been implemented?

core/src/main/java/org/jobrunr/server/tasks/startup/CheckIfAllJobsExistTask.java
```diff
@@ -47,10 +51,21 @@ public void run() {
...
-        return storageProvider.getRecurringJobs().stream()
-                .map(AbstractJob::getJobSignature)
-                .filter(jobSignature -> !jobExists(jobSignature))
-                .collect(toSet());
+        Set<String> missingRecurringJobSignatures = new HashSet<>();
+        for (RecurringJob recurringJob : storageProvider.getRecurringJobs()) {
+            if (ANNOTATION.equals(recurringJob.getCreatedBy())) {
+                if (!jobExists(recurringJob.getJobSignature())) {
+                    storageProvider.deleteRecurringJob(recurringJob.getId());
+                    LOGGER.info("Deleting recurring job {} ({}) as it was created by the @Recurring annotation but does not exist anymore", recurringJob.getId(), recurringJob.getJobSignature());
```
auloin, line 59: Should the log be above `storageProvider.deleteRecurringJob(recurringJob.getId());`? Or changed to

```suggestion
                    LOGGER.info("Deleted recurring job {} ({}) as it was created by the @Recurring annotation but does not exist anymore", recurringJob.getId(), recurringJob.getJobSignature());
```

core/src/main/java/org/jobrunr/server/tasks/startup/CheckIfAllJobsExistTask.java
```diff
@@ -47,10 +51,21 @@ public void run() {
...
+                } else if (isNotPresent(getRecurringAnnotation(recurringJob.getJobDetails()))) {
+                    storageProvider.deleteRecurringJob(recurringJob.getId());
+                    LOGGER.info("Deleting recurring job {} ({}) as it was created by the @Recurring annotation but is not annotated by the  @Recurring annotation anymore", recurringJob.getId(), recurringJob.getJobSignature());
```
auloin, line 62: Should the log be above `storageProvider.deleteRecurringJob(recurringJob.getId());`? Or changed to

```suggestion
                    LOGGER.info("Deleted recurring job {} ({}) as it was created by the @Recurring annotation but is not annotated by the  @Recurring annotation anymore", recurringJob.getId(), recurringJob.getJobSignature());
```

auloin (changes requested):

auloin (approved):

PR #1188, opened 2025-01-16
Escape Postgres table names

SIMULATAN:
## Description
This PR fixes an issue I faced using `jobrunr:` as the tablePrefix.

So far, I only implemented this for PostgreSQL as I do not have an application on hand to test other SQL-based DBs. 

## Postgres
The naming rules in PostgreSQL are as follows:
> SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($).

However, using double quotes, it is possible to use essentially any character wanted:
> Quoted identifiers can contain any character, except the character with code zero. (To include a double quote, write two double quotes.) This allows constructing table or column names that would otherwise not be possible, such as ones containing spaces or ampersands. The length limitation still applies.

*https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS*

## MySQL
MySQL would also support using quoted identifiers.
>  Permitted characters in unquoted identifiers:
>    ASCII: [0-9,a-z,A-Z$_] (basic Latin letters, digits 0-9, dollar, underscore)
>    Extended: U+0080 .. U+FFFF 
>
> Permitted characters in quoted identifiers include the full Unicode Basic Multilingual Plane (BMP), except U+0000:
>    ASCII: U+0001 .. U+007F
>    Extended: U+0080 .. U+FFFF 

*https://dev.mysql.com/doc/refman/8.4/en/identifiers.html*

## OracleDB
Same case here.
> Every database object has a name. In a SQL statement, you represent the name of an object with a quoted identifier or a nonquoted identifier.
>    A quoted identifier begins and ends with double quotation marks ("). If you name a schema object using a quoted identifier, then you must use the double quotation marks whenever you refer to that object.
>    A nonquoted identifier is not surrounded by any punctuation.
> You can use either quoted or nonquoted identifiers to name any database object.

*https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/Database-Object-Names-and-Qualifiers.html*

## SQL Server, H2, ANSI
I couldn't find any docs mentioning *rules* and *restrictions* (and thus quoting), only conventions.

### DIFF
--- a/core/src/main/java/org/jobrunr/storage/sql/common/DatabaseCreator.java
+++ b/core/src/main/java/org/jobrunr/storage/sql/common/DatabaseCreator.java
@@ -5,11 +5,7 @@ import org.jobrunr.storage.StorageException;
 import org.jobrunr.storage.sql.SqlStorageProvider;
 import org.jobrunr.storage.sql.common.db.Transaction;
 import org.jobrunr.storage.sql.common.migrations.SqlMigration;
-import org.jobrunr.storage.sql.common.tables.AnsiDatabaseTablePrefixStatementUpdater;
-import org.jobrunr.storage.sql.common.tables.NoOpTablePrefixStatementUpdater;
-import org.jobrunr.storage.sql.common.tables.OracleAndDB2TablePrefixStatementUpdater;
-import org.jobrunr.storage.sql.common.tables.SqlServerDatabaseTablePrefixStatementUpdater;
-import org.jobrunr.storage.sql.common.tables.TablePrefixStatementUpdater;
+import org.jobrunr.storage.sql.common.tables.*;
 import org.jobrunr.utils.StringUtils;
 import org.jobrunr.utils.annotations.VisibleFor;
 import org.slf4j.Logger;
@@ -126,7 +122,10 @@ public class DatabaseCreator {
     }
 
     private boolean isMigrationsTableMissing() {
-        String migrationsFQTableName = tablePrefixStatementUpdater.getFQTableName("JOBRUNR_MIGRATIONS").toUpperCase();
+        String migrationsFQTableName = tablePrefixStatementUpdater.getFQTableName("JOBRUNR_MIGRATIONS")
+            // the table name may be escaped => remove surrounding quotes
+            .replaceAll("^\"(.*)\"$", "$1")
+            .toUpperCase();
         return getAllTableNames().stream().map(String::toUpperCase).noneMatch(x -> x.contains(migrationsFQTableName));
     }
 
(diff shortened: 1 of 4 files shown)


rdehuyss:
@cla-bot check

SIMULATAN:
It appears some tests are failing as my quotation changes break schema specifications. I'll probably just split off at the first dot in table names, treat the left part as the schema and just quote the rest. Sadly, this doesn't support schemas including a dot, but I doubt that's a common thing people would like to do.

rdehuyss:
Hi @SIMULATAN - thanks for this PR!

Before we consider it for merging, we would like to know the exact reason you're creating it. JobRunr by default already prefixes it's tables with `jobrunr_`. I noticed you're using `jobrunr:` with a colon but in all my 20 years as a developer, I've never seen somebody using a colon in table names.

Can you share the reason you want to prefix it with `jobrunr:`?

SIMULATAN:
Sure. For context, I am developing a low-code automation platform organized into Nodes and Flows as my diploma thesis. In order to support niche services, "extensions" can be loaded in at runtime. As I want to avoid conflicts, they have a globally unique `namespace` associated which, for the ease of use, is the extension's package name.

To ease management, I store all business, 3rd party extension and JobRunr records in a single database.
In order to visually separate those tables (and make it easier to filter logs, group metrics by their system, ...), I came up with the colon as a distinct and rarely used character.
In practice, this looks something along those lines:
```
flows
nodes
nodes:my.third.party.package:my_node_name:service_table
nodes:my.builtin.nodes:commonly_used_node:service_table
jobrunr:migrations # actually called jobrunr:jobrunr_migrations due to the `elementPrefixer` concating the custom prefix with the default prefix
jobrunr:jobs
...
```

rdehuyss:
What about using the `$` sign? This is what we've seen used in the past for this?

If you'd like to see this PR merged, we also require specific test cases for every feature added to JobRunr (where we reference the issue with an annotation `@Because("https://github.com/jobrunr/jobrunr/pull/1188")` so we can keep track on why certain changes were done.

SIMULATAN:
Oh wow, that does work indeed. Must've missed it because I accidentally opened the PG7 docs initially.
I think the dollar sign works well enough for my use case. The "escaping all names" part is done and implemented in this PR, adding tests wouldn't be a problem either. But, as mentioned, there's still the schema issues, which greatly complicate the "regexing" and also lead to me discovering some other issues.

I think I'll close this PR for now, in case this is wanted in the future it can be used as a reference. Thank you for your considerations and time.

rdehuyss:
Good luck with the thesis!

PR #1207, opened 2025-02-19
Support kotlinx.serialization

### LINKED ISSUE #960: [Feature] add support for kotlinx.serialization as JsonMapper
**Is your feature request related to a problem? Please describe.**
With [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization) gaining popularity in the Kotlin space, it would be great to have it as a `JsonMapper` option in JobRunr and avoid the need to have different type of json libraries in the same project.

**Describe the solution you'd like**
A configurable `JsonMapper` that uses `kotlinx.serialization` is added to JobRunr `core` or `language-support`.

**Describe alternatives you've considered**
The only alternative would be to develop it as a separate project, independent of JobRunr.

**Additional context**
If it's okay to have it in JobRunr, I'm open to creating a PR for it.

SIMULATAN:
This PR implements a fully featured `JsonMapper` backed entirely by kotlinx.serialization.

Closes #960 

## Current State
I am proposing the current version to gather feedback to hopefully get this PR ready soon-ish.
Unfortunately, as JobRunr is written in Java and *very* strict about the expected output, I had to write very single serializer by hand. This 12-hour process (!) was not pretty and is not guaranteed to be 100% correct, therefore...

## Testing
If you can, PLEASE test the new mapper yourself!

I published the current release (`7.4.0-kotlinx.1`) to [my maven repository](https://maven.simulatan.me/#/snapshots/org/jobrunr).
You can use it in your project by adding the `https://maven.simulatan.me/snapshots` repository and updating all jobrunr dependencies to use the aforementioned version (make sure to also use my `jobrunr-core` version as some visibilities were changed!).
Sadly, I haven't gotten Drone running yet, therefore, you'll just have to trust me or build the library from source.

I verified the new implementation myself using my real application locally and will conduct tests in my staging environment tomorrow.

<details><summary>Screenshots</summary>

dashboard
job overview
single job

</details> 

### Usage Notes
- Of course, the usual "**no warranty, back everything up before trying, not production ready**" clause applies here as well.
- Simply replace your previous `JsonMapper` with `KotlinxSerializationJsonMapper`
	- [JobRunrConfiguration#useJsonMapper](https://javadoc.io/doc/org.jobrunr/jobrunr/latest/org/jobrunr/configuration/JobRunrConfiguration.html#useJsonMapper(org.jobrunr.utils.mapper.JsonMapper))
- All job requests must be serializable! The respective error message is somewhat cryptic right now..
- Please report all serialization related issues immediately!
	- Exceptions thrown during the serialization process are currently logged directly before propagating through the system to make sure no problems go unnoticed

## Things to discuss
- [x] had to use the verification JSONs for the `JsonMapperTest`, therefore, I moved all of them to the `testFixtures` sourceSet
      => do we want to keep them all there?
- [ ] should the `kotlinx-serialization-json` library be `compileOnly`
      => no other implementations can be used, therefore, it'd make sense to use `implementation` instead
- [ ] `@OptIn` to `(Internal|Experimental)SerializationApi` on the module level using Gradle?
- [ ] what fields are incorrectly assumed to be nonnull?
- [ ] general review, code style, code smell, etc.


core/src/main/java/org/jobrunr/jobs/context/JobDashboardLogger.java
```diff
@@ -64,6 +65,10 @@ public JobDashboardLogLines() {
         public void add(JobDashboardLogLine line) {
             logLines.add(line);
         }
+        
+        public void addAll(Collection<JobDashboardLogLine> lines) {
+            logLines.addAll(lines);
+        }
```
SIMULATAN, line 71: Figured this approach was in-line with the existing `add`.
Alternatively, a secondary constructor taking an initial collection could be added.

core/src/main/java/org/jobrunr/jobs/context/JobDashboardLogger.java
```diff
@@ -86,6 +91,12 @@ public JobDashboardLogLine(Level level, String logMessage) {
             this.logMessage = logMessage;
         }
 
+        public JobDashboardLogLine(Level level, Instant logInstant, String logMessage) {
+            this.level = level;
+            this.logInstant = logInstant;
```
SIMULATAN, line 96: `logInstant` is set to `Instant.now()` in the other constructor, other `JsonMapper`s use reflection hacks to override it

core/src/main/java/org/jobrunr/jobs/context/JobDashboardProgressBar.java
```diff
@@ -81,7 +81,7 @@ private static String progressBarKey(int jobStateNbr) {
         return JOBRUNR_PROGRESSBAR_KEY + "-" + jobStateNbr;
     }
 
-    private static class JobDashboardProgress implements JobContext.Metadata {
+    public static class JobDashboardProgress implements JobContext.Metadata {
```
SIMULATAN, line 84: needed to reference it in the new serializer

core/src/main/java/org/jobrunr/jobs/states/AbstractJobState.java
```diff
@@ -23,6 +23,10 @@ public Instant getCreatedAt() {
         return createdAt;
     }
 
+    public void setCreatedAt(Instant createdAt) {
+        this.createdAt = createdAt;
+    }
```
SIMULATAN, line 28: otherwise only setable using reflection

core/src/main/java/org/jobrunr/jobs/states/FailedState.java
```diff
@@ -40,6 +40,17 @@ public FailedState(String message, Exception exception) {
...
+    public FailedState(String message, String exceptionType, String exceptionMessage, String exceptionCauseType, String exceptionCauseMessage, String stackTrace, boolean doNotRetry) {
+        super(StateName.FAILED);
+        this.message = message;
+        this.exceptionType = exceptionType;
+        this.exceptionMessage = exceptionMessage;
+        this.exceptionCauseType = exceptionCauseType;
+        this.exceptionCauseMessage = exceptionCauseMessage;
+        this.stackTrace = stackTrace;
+        this.doNotRetry = doNotRetry;
+    }
```
SIMULATAN, line 52: The other available "non-no-arg-constructor" eats an `Exception`. I also tried making the restore function `public static` but figured it's more reasonable to skip the restore-and-extract step entirely.

core/src/testFixtures/java/org/jobrunr/JobRunrAssertions.java
```diff
@@ -87,8 +86,10 @@ public static DatabaseAssertions assertThat(DataSource dataSource) {
     }
 
     public static String contentOfResource(String resourceName) {
-        try {
-            return Files.readString(Paths.get(JobRunrAssertions.class.getResource(resourceName).toURI()));
+        try (InputStream inputStream = JobRunrAssertions.class.getResourceAsStream(resourceName)) {
```
SIMULATAN, line 89: the old approach failed after moving the resources to the `testFixtures` source set

rdehuyss:
Hi Jakob! Nice work on this! We'll have a look next week - thanks for your hard work!

language-support/jobrunr-kotlin-21-support/src/main/kotlin/org/jobrunr/kotlin/utils/mapper/KotlinxSerializationJsonMapper.kt
```diff
@@ -0,0 +1,146 @@
...
+
+		this.serializersModule = serializersModule + jobRunrSerializersModule
+	})
+
+	constructor() : this(jobRunrSerializersModule)
+
+	override fun serialize(obj: Any): String? = rethrowSerializationException {
+		fun <T : Any> encode(obj: Any): String? {
+			return json.encodeToString(
+				json.serializersModule.serializer(obj::class) as SerializationStrategy<T>?
```
wgroeneveld, line 103: As mentioned before, for many POJOs this still returns `null` resulting in a `noSerializerFound`: another example is `JobParameter` while `JobStats` works fine

wgroeneveld (commented):
Thanks a lot for your hard work! I think we need some more modifications before this can be ready. 

The biggest thing I am worried about though is much increased boilerplate that's not kept in sync with the actual code: the serializers themselves that list all (but sometimes not!) fields using `Field("name", ...)`. 
I wonder whether we can create a **fitness test to validate this** against the actual POJO to make sure we never forget a property or if we do we do it on purpose. 
I also wonder whether we really need all those things? I know kotlinx serialization does not use reflection but with setters/constructors set properly we would technically not need alll those serializers.

So perhaps we're trying to fix this "wrong" and should fix it at the root cause, namely the POJOs themselves? 
What do you think?

SIMULATAN:
Hi, I apologize for the delay, 'been busy with exams.

> The biggest thing I am worried about though is much increased boilerplate that's not kept in sync with the actual code

Yep, that was also my biggest pain point with this implementation. Usually, you'd write entirely manual serialization logic, therefore duplicating a good few lines. To address this, I decided to abstract these things into field-based serializers.

>  I also wonder whether we really need all those things? I know kotlinx serialization does not use reflection but with setters/constructors set properly we would technically not need alll those serializers.

Unfortunately, kotlinx.serialization expects certain elements in the bytecode that are only present in files compiled from Kotlin code (such as default parameter values to deserialize objects). Unfortunately, kotlinx.serialization only generates unified serializers and deserializers, therefore breaking when no setters are available. Even in manual serializers, I sometimes have to use setters to alter an instantiated POJO to get it into the correct state.
Here are some examples for error messages thrown when attempting to let kotlinx.serialization generate serializers. As evident, it is not set up to handle Java classes and simply thorws internal error messages (in fact, the omitted log contains a plea to "report this problem" to the Kotlin bug tracker)
https://gist.github.com/SIMULATAN/497a9db2ad63e61a2e5dbf8726b2293c

> So perhaps we're trying to fix this "wrong" and should fix it at the root cause, namely the POJOs themselves?

As a huge Kotlin fan myself, I would of course love to rewrite the models in Kotlin. As an example, the `JobStats` model would greatly benefit from the Kotlin Data Class treatment, we could cut a *ton* of lines. Whilst we're at it, standardizing construction and access of classes is probably also a good idea. Not sure if this is wanted or entirely out of scope though..?


I will address the individual review comments in a second.

language-support/jobrunr-kotlin-21-support/src/test/java/org/jobrunr/jobs/mappers/KotlinxSerializationJobMapperTest.java
```diff
@@ -0,0 +1,11 @@
+package org.jobrunr.jobs.mappers;
+
+import org.jobrunr.kotlin.utils.mapper.KotlinxSerializationJsonMapper;
+import org.jobrunr.utils.mapper.JsonMapper;
+
+public class KotlinxSerializationJobMapperTest extends JobMapperTest {
+	@Override
+	protected JsonMapper getJsonMapper() {
+		return new KotlinxSerializationJsonMapper(TestKotlinxSerializationModuleKt.getTestModule());
```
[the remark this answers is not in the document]
SIMULATAN: > For me this test consistently fails:

That's curious, I've had this problem in earlier commits where I serialized `Instant`s using the epoch timestamp, but it was gone once I switched to the ISO-8601 based serialization.
Haven't faced this kind of errors ever since, and the CI also passes consistently..

> If there are tests missing, I think it is our responsibility and not that of @SIMULATAN .

I'd greatly appreciate if these kind of tests could be provided upstream, I could then rebase and make sure the JsonMapper implementation conforms to the spec. Due to time and knowledge constraints and to avoid misalignment, I'd refrain from writing complex mapping tests myself.

language-support/jobrunr-kotlin-21-support/src/main/kotlin/org/jobrunr/kotlin/serialization/utils/FieldBasedSerializer.kt
```diff
@@ -0,0 +1,51 @@
...
+				val fieldValue = field.getter(value) as O?
+				if (field.nullable) encodeNullableSerializableElement(descriptor, index, serializer, fieldValue)
+				else encodeSerializableElement(descriptor, index, serializer, fieldValue ?: error("Field ${field.name} is null"))
+			}
+
+			serialize<Any>()
+		}
+	}
+	
+	override fun deserialize(decoder: Decoder): T = throw DeserializationUnsupportedException()
```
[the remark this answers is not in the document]
SIMULATAN: Although I planned on supporting deserialization for every entity, I decided to drop this strategy for all UI-related entities. Since they are merely used to supply information to the Frontend, I don't think deserialization is necessary?
Unfortunately, most of these entities don't even seem to support external constructions as they lack public constructors or setters, further confirming my assumption.

language-support/jobrunr-kotlin-21-support/src/main/kotlin/org/jobrunr/kotlin/serialization/jobs/states/JobStateSerializer.kt
```diff
@@ -0,0 +1,185 @@
...
+				}
+			}
+		}
+		return createdAt
+	}
+}
+
+object DeletedStateSerializer : JobStateSerializer<DeletedState>(
+	DeletedState::class,
+	Field("reason", String.serializer()) { it.reason },
```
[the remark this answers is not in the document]
SIMULATAN: `DeletedStateSerializer` uses the `JobStateSerializer` base class. It has two default fields that it serializers for every `JobState` implementation: `state` and `createdAt`. The `reason` is an additional field specific to the `Deleted` state.
Therefore, all other `AbstractJobState`s are also covered and (should) be serializable and deserializable.

language-support/jobrunr-kotlin-21-support/src/main/kotlin/org/jobrunr/kotlin/utils/mapper/KotlinxSerializationJsonMapper.kt
SIMULATAN: `JobParameter` should be serializable and deserializable using the [JobParameterSerializer](https://github.com/SnotyMe/jobrunr/blob/5b1800a209b3c3d8ddf3c7d7979004639ba712bb/language-support/jobrunr-kotlin-21-support/src/main/kotlin/org/jobrunr/kotlin/serialization/jobs/JobParameterSerializer.kt). I've tried running my application (which utilizes `JobParameter`s) using my branch and it worked just fine.
However, yes, chances are some POJOs are not covered yet as I just haven't encountered them whilst randomly clicking around the dashboard.

language-support/jobrunr-kotlin-21-support/src/main/kotlin/org/jobrunr/kotlin/serialization/utils/FieldBasedSerializer.kt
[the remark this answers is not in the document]
wgroeneveld: Ok check, thanks for the clarification I think that's fine if not all UI-specific objects can get deserialized again, there are probably corner cases for jsonB as well that are left uncovered.

core/src/main/java/org/jobrunr/jobs/context/JobDashboardLogger.java
rdehuyss: Let's go for the secondary constructor approach? I think it makes more sense in the long run.

core/src/main/java/org/jobrunr/jobs/context/JobDashboardLogger.java
rdehuyss: Makes sense. Let's have the other constructor call this constructor however to minimize code.

core/src/main/java/org/jobrunr/jobs/states/FailedState.java
rdehuyss: Looks good to me! I would only make sure the other constructors use this one.

rdehuyss:
Hi @SIMULATAN - we're currently looking into this but as none of us our Kotlin experts this is a difficult PR for us to review hence the delays 😳. As any question related to this will come to us (e.g. a possible issue) we want to make sure we do this right. So, sorry about the delays on our side 🫣.

I'm also wondering if @gmariotti and @alexanderpichler-cpi have some time to review this?

That being said, my 5 cents:
- For now, I think rewriting the models entirely to Kotlin is a bit to soon but I indeed prefer to have as much as possible immutable. I think we can add the necessary constructors to the different JobRunr classes to have a public constructor where necessary.
- I would also like to move all tests related to the dashboard to `testFixtures` and have the tests related to the dashboard (`JobRunrDashboardWebServerTest`) in `testFixtures` and also have a `KotlinXJobRunrDashboardWebServerTest`

We will add upstream tests (can we push to your branch?) for tests we think are still missing.

Last question: are you ok with this being in JobRunr v8 and we explicitly give you credit in the release of JobRunr v8 and during our webinar?

SIMULATAN:
> As any question related to this will come to us (e.g. a possible issue) we want to make sure we do this right.

Fair. Just so you know, I'm always here to assist, especially with the Kotlin integration, and am willing to maintain the new JsonMapper in the foreseeable future.

> So, sorry about the delays on our side

No worries. I'd appreciate it if we could get it merged sometime in March (as April is my project's deadline), but since Jackson is working just fine as-is, there's no need to rush things.

> For now, I think rewriting the models entirely to Kotlin is a bit to soon but I indeed prefer to have as much as possible immutable.

Makes sense. I'll push the changes from the review comments in a sec. I'd love to contribute to any potential Kotlin-related efforts, feel free to ping me when those things come up.

> can we push to your branch?

I can't find the checkbox, but you should be able to. In any case, I can always rebase or cherry-pick from main or any other branch.


In general, I can reason with the feedback and think we're on the right path.

> Last question: are you ok with this being in JobRunr v8 and we explicitly give you credit in the release of JobRunr v8 and during our webinar?

I'd love it to be! My only requirement is that my JsonMapper module must stay Open Source (as I cannot afford a Pro license and would like my contribution to benefit the general public).

core/src/main/java/org/jobrunr/jobs/states/FailedState.java
SIMULATAN: While we're at it, it may be worth caching the reconstructed exception built in the `getException` method by updating the `exception` field. Is there any reason why this isn't done already?

language-support/jobrunr-kotlin-21-support/src/test/java/org/jobrunr/jobs/mappers/KotlinxSerializationJobMapperTest.java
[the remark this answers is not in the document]
SIMULATAN: Just ran the `check` task again and faced this problem. Definitely some flakiness going on, but I have zero clue where it could be. Serialization and deserialization uses the ISO-8601 format and therefore should be perfectly accurate. In my case, it clearly isn't rounding or truncation either...
```java
field/property 'jobHistory[0].createdAt' differ:
- actual value  : 2025-03-06T13:39:06.682401562Z
- expected value: 2025-03-06T13:39:06.668136389Z
Compared objects have java types and were thus compared with equals method
field/property 'jobHistory[1].createdAt' differ:
- actual value  : 2025-03-06T13:39:06.682657817Z
- expected value: 2025-03-06T13:39:06.668249085Z
Compared objects have java types and were thus compared with equals method
field/property 'jobHistory[2].createdAt' differ:
- actual value  : 2025-03-06T13:39:06.682993825Z
- expected value: 2025-03-06T13:39:06.668377503Z
Compared objects have java types and were thus compared with equals method
field/property 'jobHistory[3].createdAt' differ:
- actual value  : 2025-03-06T13:39:06.683078549Z
- expected value: 2025-03-06T13:39:06.668380338Z
Compared objects have java types and were thus compared with equals method
field/property 'jobHistory[4].createdAt' differ:
- actual value  : 2025-03-06T13:39:06.683577086Z
- expected value: 2025-03-06T13:39:06.668425800Z
```

Since I cannot reproduce the issue when just running the test class on its own, I suspect a timing issue caused by some `Instant.now()` somewhere. I'm confused, may have to debug further.

SIMULATAN:
Rebased onto v8, will push fixes and address the review comments in fixup commits soon

auloin:
Hi @SIMULATAN, thanks for the great work so far. I think our main concern at the moment is readability/complexity of the code (and some extra tests needed but this is more for us to figure out). As far as I can tell, there is not much we can do about the code complexity. So I made an attempt at making it more readable.

Can I ask for your opinion on this attempt? From my, java developer, perspective I find this version easier to read but it may not be the case for a Kotlin developer (it may not even follow Kotlin/kotlinx.serialization standards). You can find my changes at https://github.com/auloin/jobrunr-kotlinx-serialization/commit/fab46f256f997982a688d55688aa7c177ae1059e, I've mainly only reworked the `JobStateSerializer` file.

> EDIT: forgot to ask, what is your preferred merging strategy? Not sure how I'd pack my changes into atomic commits.

We squash and merge PRs, so we don't enforce commit standards at PR level.

rdehuyss:
@SIMULATAN : another FYI - we're reaching out to friends in the kotlin community who know Kotlin more than we do to have a look at this PR.

SIMULATAN:
@auloin I can see the idea. In the beginning, I did consider creating `K*` "copies" too as they'd remove the need for manual serializers and therefore greatly reduce the maintenance overhead. As far as I can remember, the only reason I decided against it was to prevent duplicating work and having multiple models to keep in sync. In that case, I should be able to bring back [the previously removed `KSerializable` approach](https://github.com/jobrunr/jobrunr/pull/1207/commits/8ab00794706999359cdb302ce593d2cc0dcbf248#diff-a3a4b2045373b665fc8866b5ab9fced92b59ec3b70eef2621782dc5dd0f34788).
Skimming over the commit, I believe there's even some potential left to massively reduce the boilerplate of your state serializers. I will try my best to apply the changes manually in the coming days, no guarantees though as I'm right in the most stressful weeks of the year.

> We squash and merge PRs, so we don't enforce commit standards within a PR level.

Ok, that's good. I'll refrain from manually squashing or rebasing in the future then, that'll make it significantly easier to keep track of changes.

@rdehuyss 
> another FYI - we're reaching out to friends in the kotlin community who know Kotlin more than we do to have a look at this PR.

Great to hear, looking forward to getting some input from them too. I'm hoping to get these changes out in time.

SIMULATAN:
Pushed the proposed changes. They should work, I briefly tested scheduling and the dashboard.

I'm currently looking into the CI failure by running the full `check` task locally.
Sadly, the `RecurringJob` serializer will probably have to stay a manual serializer as the `RecurringJobUIModel` serialization [depends on it](https://github.com/jobrunr/jobrunr/pull/1207/files#diff-8bc2b58a8df5e2e33ed8a469efbc5295bba6f3e69c18441ad5adbcd5752b3d06).

rdehuyss:
The CI failure is probably not your fault. I restarted the build, let's see.

rdehuyss:
For me, this looks really good and it is almost ready for merging. On our macs, we still have one failing test that I want to look into.

For the rest, excellent work!

rdehuyss:
I tried to fix the test failing on our side but I cannot push to your branch:
15:12:15.667: [jobrunr] git -c credential.helper= -c core.quotepath=false -c log.showSignature=false push --progress --porcelain SnotyMe refs/heads/fork/SnotyMe/feat/960-kotlinx-serialization:feat/960-kotlinx-serialization
ERROR: Permission to SnotyMe/jobrunr.git denied to rdehuyss.
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.

SIMULATAN:
Turns out the toggle is not available when the fork belongs to an organization... I have invited you to the repository instead. Alternatively, I can cherry-pick your commit from another branch of this repo.

https://github.com/orgs/community/discussions/5634

auloin (changes requested):
@SIMULATAN we're happy with the PR but I'd like to request two changes:
1. Could you please revert the changes to `.idea`.
2. We want to test the serializer against the following existing tests: `AbstractJsonMapperTest`, and `JobRunrDashboardWebServerTest`. Is this something you want/have time to look into?

edit: feel free to update the state of the PR to ready 🎉

.idea/.name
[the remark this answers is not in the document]
SIMULATAN: Looks like those were removed during the rebase onto v8.
I `git checkout upstream/v8 -- .idea` and committed these files (4e83eb4a)

SIMULATAN:
> We want to test the serializer against the following existing tests: AbstractJsonMapperTest, and JobRunrDashboardWebServerTest. Is this something you want/have time to look into?

Absolutely, not sure how I missed those specs...

Sadly, there are still 3 failing tests which are due to deserializations of `Any` values. In my opinion, there's some conceptual questions left unanswered. For example, the `object` property of the `JobParameter` is only serializable by using the two class names that are persisted alongside the generic object. So far, so logical, but unfortunately, when dealing with lists, we're facing a whole different story. Since generics are erased in the serialization process, it is entirely unclear which class to deserialize the elements to. For example, the `AbstractJsonMapperTest#testCanSerializeCollections` test uses a `Long`. I did try [rewriting the AnyInlineSerializer to simply try all primitives](https://gist.github.com/SIMULATAN/ff56b27af09727065a0511a3d2419e78) to at least pass the test, but that didn't work out because the lexer just stays at the last position rather than backtracking when it encounters issues. For example, when trying a String first, it'll fail due to the missing quote but not backtrack. Therefore, all consecutive attempts to parse, say, a number, will fail as the cursor has already skipped over the start of the element. When the `object` is simply `[1]`, it'll skip to the closing bracket. Even if this approach worked, it would fail in practice as it is impossible to know what the user wants when data types overlap (like Float and Double).

Another major issue is the `AbstractJsonMapperTest#testSerializeAndDeserializeEnqueuedJobComingFrom4Dot0Dot0`. A different format is used there which uses an array with two different data types. I managed to solve this problem using a dedicated serializer, but, as it is impossible to distinquish between the cases without backtracking, deciding when to use it is impossible.

While waiting for further input, I will begin the `JobRunrDashboardWebServerTest` implementation.

auloin:
@SIMULATAN the changes at https://github.com/auloin/jobrunr-kotlinx-serialization/commit/88d9a07f128bc026a1e97caee069d143893668d9 seem to fix the tests. 

There were other tests failing for me, all related to `JobStateSerializer`; the changes I made to those serializers fix all the tests.

> Since generics are erased in the serialization process, it is entirely unclear which class to deserialize the elements to.

JobRunr already saves all the info needed to deserialise the `Set` (see below, each item has a type attached), so using the `AnySerializer` to deserialize the `JobParameterSerializer#object` seems to do the trick. However I'm confused by `AnyInlineSerializer`, it doesn't implement `deserialize`. Why do we need both?

```
"jobParameters" : [ {
  "className" : "java.util.Set",
  "actualClassName" : "java.util.Collections$SingletonSet",
  "object" : [ {
    "@class" : "java.lang.Long",
    "value" : 1
  } ]
} ]
```

> Another major issue is the `AbstractJsonMapperTest#testSerializeAndDeserializeEnqueuedJobComingFrom4Dot0Dot0`

I propose we override the test and leave it blank. It's unlikely anyone would migrate from v4.0.0, released more than 3 years ago, to directly using the kotlin serialization in v8. WDYT @rdehuyss?

---

@rdehuyss it looks like we turned `canSerializeAndDeserializeJobWithAllStatesAndMetadata` into a flaky test since the resolution pre and post deserialisation are now different.

```
field/property 'jobHistory[4].processDuration' differ:
- actual value  : 0.001448999S
- expected value: 0.001449S
```

Why do we need a custom serializer for durations? Can't we simply rely on `Duration.toString()`/`Duration.parse()`? Also using the [builtin kotlinx.serialization Duration serializer](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/builtin-classes.md#duration) makes the test green.

rdehuyss:
> @rdehuyss it looks like we turned `canSerializeAndDeserializeJobWithAllStatesAndMetadata` into a flaky test since the resolution pre and post deserialisation are now different.
> 
> ```
> field/property 'jobHistory[4].processDuration' differ:
> - actual value  : 0.001448999S
> - expected value: 0.001449S
> ```
> 
> Why do we need a custom serializer for durations? Can't we simply rely on `Duration.toString()`/`Duration.parse()`? Also using the [builtin kotlinx.serialization Duration serializer](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/builtin-classes.md#duration) makes the test green.

The frontend expects a Duration in seconds and nanos and does not support a Java Duration. I was able to fix the flaky test: the cause was that we first created a `double` and then a `BigDecimal` but the double of course has rounding errors.

auloin (approved):

auloin:
@SIMULATAN we're ready to merge the PR 🎉 Thanks for the great work and your patience.

PR #1229, opened 2025-04-02
cleanup creation of datasources in storage provider tests

wgroeneveld:
- cleanup duplication of datasources
- make `getDataSource` public to better comply additional tests in Pro


rdehuyss (approved):
LGTM!

PR #1232, opened 2025-04-04
Add architecture test to ensure classes named Abstract have abstract modifier

wgroeneveld:
This PR:

- Includes an arch test to validate classes named "Abstract" should have the modifier `abstract`
- Fixes the `ForAllSubclassExtension` that does not take into account abstract test classes that are not abstract.


core/src/test/java/org/jobrunr/architecture/ClassNamesTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.junit.jupiter.extension.ForAllSubclassesExtension;
+
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
+
+@AnalyzeClasses(packages = "org.jobrunr", importOptions = {ImportOption.DoNotIncludeTests.class, PackageDependenciesTest.DoNotIncludeTestFixtures.class})
+public class ClassNamesTest {
+
+    @ArchTest
+    ArchRule classesNamedAbstractShouldBeAbstractAsWell = classes()
+            .that().haveSimpleNameContaining("Abstract")
```
auloin, line 23: Shouldn't this be startWith instead of contain?

auloin (approved):

core/src/test/java/org/jobrunr/architecture/ClassNamesTest.java
wgroeneveld: Yes! Didn't know that existed as well!

PR #1233, opened 2025-04-04
Add and test existing 404s for the Webserver API

wgroeneveld:
This PR verifies `JobNotFoundException` as proper HTTP response code 404s by adding additional tests for the webserver API and fixes a specific case where the recurring job is not found but does not return the correct http response code.


core/src/main/java/org/jobrunr/storage/InMemoryStorageProvider.java
```diff
@@ -276,8 +276,8 @@ public boolean recurringJobsUpdated(Long recurringJobsUpdatedHash) {
 
     @Override
     public int deleteRecurringJob(String id) {
-        recurringJobs.removeIf(job -> id.equals(job.getId()));
-        return 0;
+        boolean removed = recurringJobs.removeIf(job -> id.equals(job.getId()));
+        return removed ? 1 : 0;
```
auloin, line 280: I'm surprised by this; do we have a storage provider test covering it?

core/src/main/java/org/jobrunr/dashboard/JobRunrApiHandler.java
```diff
@@ -109,7 +109,10 @@ private HttpRequestHandler getRecurringJobs() {
 
     private HttpRequestHandler deleteRecurringJob() {
         return (request, response) -> {
-            storageProvider.deleteRecurringJob(request.param(":id"));
+            int deleted = storageProvider.deleteRecurringJob(request.param(":id"));
+            if(deleted == 0) {
+                response.statusCode(404);
```
auloin, line 114: Should this a throw a `JobNotFoundException` and we let the exception handler what to do with it?

core/src/main/java/org/jobrunr/dashboard/JobRunrApiHandler.java
```diff
@@ -109,7 +109,10 @@ private HttpRequestHandler getRecurringJobs() {
...
+            }
             response.statusCode(204);
```
auloin, line 116: If we keep this as is; shouldn't this line be in an else statement?

auloin (changes requested):

core/src/main/java/org/jobrunr/storage/InMemoryStorageProvider.java
wgroeneveld: There isn't a single test in the storage provider test that only tests this but `testCRUDRecurringJobLifeCycle` also deletes recurring jobs. Should we an an assertion as we don't use the result now?

auloin: > Should we an an assertion as we don't use the result now?

I think this makes sense

auloin (approved):

auloin (approved):

PR #1255, opened 2025-05-07
Reduce the amount of methods exposed by JobScheduler and JobRequestScheduler

auloin:


core/src/test/java/org/jobrunr/scheduling/AbstractJobSchedulerTest.java
```diff
@@ -17,6 +32,42 @@
 
 class AbstractJobSchedulerTest {
 
+    @Test
+    void scheduleValidatesTemporalType() {
+        var storageProvider = new InMemoryStorageProvider();
+        storageProvider.setJobMapper(new JobMapper(new JacksonJsonMapper()));
+        AbstractJobScheduler jobScheduler = jobScheduler(storageProvider);
```
rdehuyss, line 39: I don't know whether I would add this test as it is covered by InstantUtils where it makes most sense. Perhaps only one to make sure we use the InstantUtils?

rdehuyss (approved):
Awesome!

PR #1258, opened 2025-05-09
Drop `org` prefix from JobRunr's Spring properties

auloin:
Breaking:

Change property prefix from `org.jobrunr` to `jobrunr`. Example:

`org.jobrunr.background-job-server.enabled=true
` becomes `jobrunr.background-job-server.enabled=true`


rdehuyss (approved):
LGTM!

PR #1309, opened 2025-06-24
Add notification center

auloin:


core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/notifications/carbon-intensity-api-error-problem.js
```diff
@@ -0,0 +1,22 @@
...
+        <DismissibleClusterProblemNotification
+            endpoint="/api/problems/carbon-intensity-api-error"
+            severity="warning"
+            title="Carbon Aware API Warning"
+            date={problem.createdAt}
+            read={problem.read}
+            {...rest}
+        >
+            <p>
+                The Background Job Server has carbon aware jobs enabled but the Carbon Intensity API is currently unreachable. Please make sure the API is
```
brainbaking, line 14: I presume here is where we need to add more info coming from he problem originating at the job task

```java
            if (carbonIntensityForecast.hasError()) {
                dashboardNotificationManager.notify(new CarbonIntensityApiErrorNotification());
            }
```

where we then pass in the forecast apiResponse code + message (which contains the exception message)?
(this is issue https://github.com/orgs/jobrunr/projects/4/views/1?pane=issue&itemId=116541118 then)

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/notifications/new-jobrunr-version-available.js
```diff
@@ -8,6 +8,7 @@ const getVersionParts = (version) => {
 }
 
 const versionIsNewerThanOther = (version, otherVersion) => {
+    // TODO ignore betas
```
brainbaking: Maybe we don't need to do this?

auloin: Why? The notification for newer versions says to updgrade asap, I worry about some users eagerly upgrading without considering the beta status.

Or should we adapt the message depending on the tag?

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/notifications/severe-jobrunr-exception-problem.js
```diff
@@ -42,12 +44,19 @@ const SevereJobRunrExceptionProblem = ({problem, hasCpuAllocationIrregularity, r
...
-            title="Fatal"
+        <DismissibleClusterProblemNotification
+            title="Severe JobRunr Exception"
             severity="error"
             endpoint="/api/problems/severe-jobrunr-exception"
-            refresh={refresh}
-            extraAction={<Button variant="outlined" onClick={toggleShowIssueDialog} size="small">More&nbsp;details</Button>}
+            date={problem.createdAt}
+            extraMenuItems={
+                <MenuItem onClick={toggleShowIssueDialog}>
```
brainbaking: As discussed; perhaps this is not discoverable enough and deserves a separate button right within the popup itself, not in the three dots?

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/notifications/top-app-bar-notifications-center.js
```diff
@@ -0,0 +1,278 @@
...
+    const reloadProblems = useCallback(() => {
+        fetch(`/api/problems`).then(res => res.json())
+            .then(problems => setClusterProblems(problems))
+            .catch(e => console.error("Failed to reload problems", e))
+    }, []);
+
+    useEffect(() => {
+        Promise.all([
+            fetch("/api/problems").then(res => res.json()),
+            fetch("https://api.jobrunr.io/api/version/jobrunr/latest").then(res => res.json()).catch(() => undefined /* ignored */),
```
brainbaking: Just to be sure: I presume we ignore any errors here (was already there) in case people block the outgoing call and we don't want to make a fuss/notice?

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/notifications/new-jobrunr-version-available.js
brainbaking: As in make it clear that the release is a beta one **in bold** or with a label or something?

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/notifications/top-app-bar-notifications-center.js
auloin: Exactly; also before this would have blocked the entire UI (was on the entry component Admin.js)

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/notifications/new-jobrunr-version-available.js
auloin: As in not conveying the same urgency to upgrade as we'd do for a stable release. 

Something like :

```
isStableVersion ? "Upgrade as soon as possible to enjoy the new features, performance improvements and bug fixes" : "We've released {version} and are looking for feedback. You start a discussion over on GitHub. Thanks for your contribution!"
```

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/notifications/severe-jobrunr-exception-problem.js
auloin: Done

PR #1310, opened 2025-06-24
CarbonAware Job Processing - fix layout labeling 

wgroeneveld:
Fixes:

- The space in the job history

Adds:

- The region in the reason when scheduling a carbon aware job
- A carbon aware information popup in the job detail page when scheduling a carbon aware job

Screenshot 2025-06-24 at 15 44 30

### DIFF
--- a/core/src/main/java/org/jobrunr/server/tasks/zookeeper/ProcessCarbonAwareAwaitingJobsTask.java
+++ b/core/src/main/java/org/jobrunr/server/tasks/zookeeper/ProcessCarbonAwareAwaitingJobsTask.java
@@ -106,7 +106,7 @@ public class ProcessCarbonAwareAwaitingJobsTask extends AbstractJobZooKeeperTask
         } else if (carbonIntensityForecast.hasNoForecastForPeriod(state.getFrom(), state.getTo())) {
             scheduleJobAt(job, state.getFallbackInstant(), state, getReasonForMissingForecast(state));
         } else {
-            scheduleJobAt(job, idealMoment(state), state, "At the best moment to minimize carbon impact.");
+            scheduleJobAt(job, idealMoment(state), state, "At the best moment to minimize carbon impact in " + this.carbonIntensityForecast.getDisplayName() + ".");
         }
     }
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/jobrunr/server/tasks/zookeeper/ProcessCarbonAwareAwaitingJobsTask.java
```diff
@@ -106,7 +106,7 @@ void moveCarbonAwareJobToNextState(Job job) {
         } else if (carbonIntensityForecast.hasNoForecastForPeriod(state.getFrom(), state.getTo())) {
             scheduleJobAt(job, state.getFallbackInstant(), state, getReasonForMissingForecast(state));
         } else {
-            scheduleJobAt(job, idealMoment(state), state, "At the best moment to minimize carbon impact.");
+            scheduleJobAt(job, idealMoment(state), state, "At the best moment to minimize carbon impact in " + this.carbonIntensityForecast.getDisplayName() + ".");
```
auloin: Since this is displayed as a title for now, I'd remove the dot at the end. Usage of the dot is already inconsistent even in this method (see the other calls to `scheduleJobAt`).

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/jobs/job-view.js
```diff
@@ -188,6 +189,7 @@ const JobView = (props) => {
                             {job.jobDetails.cacheable === false && <JobDetailsNotCacheableNotification job={job}/>}
                             {stateBreadcrumb.state === 'SUCCEEDED' && <SucceededNotification job={job}/>}
                             {stateBreadcrumb.state === 'DELETED' && <DeletedNotification job={job}/>}
+                            {stateBreadcrumb.state === 'SCHEDULED' && <CarbonAwareScheduledNotification job={job}/>}
```
auloin, line 192: As we have been discussing: we should we also have this on `AWAITING` state?

auloin (approved):

PR #1313, opened 2025-06-25
add carbon intensity api response code + message in the notification centre

wgroeneveld:
Usability: CarbonIntensityApiErrorNotification should report the exception message

This will allow the user to know why an exception happened and help us diagnose issues when reported.


core/src/main/java/org/jobrunr/dashboard/ui/model/problems/CarbonIntensityApiErrorProblemHandler.java
```diff
@@ -36,7 +36,7 @@ public void onChange(List<JobRunrMetadata> metadataList) {
         if (this.metadata == null || this.metadata.size() != metadataList.size()) {
             dismiss();
             if (!metadataList.isEmpty()) {
-                problems.addProblem(new CarbonIntensityApiErrorProblem(InstantUtils.max(metadataList.stream().map(JobRunrMetadata::getCreatedAt))));
+                problems.addProblem(new CarbonIntensityApiErrorProblem(metadataList, InstantUtils.max(metadataList.stream().map(JobRunrMetadata::getCreatedAt))));
```
auloin: Since we now pass `metadataList` to  `CarbonIntensityApiErrorProblem`, we can remove its second constructor and compute `InstantUtils.max(metadataList.stream().map(JobRunrMetadata::getCreatedAt))` there. Similarly to what we do for the other problems.

PR #1322, opened 2025-06-29
Visualize carbon intensity

rdehuyss:
Initial implementation to visualize carbon.

image


wgroeneveld (approved):
Did some more tests and it's looking good!

PR #1327, opened 2025-07-03
Hotfix: v8 RecurringJob#createdBy breaks kotlinx.serialization

SIMULATAN:
In 56bc132 (due to be released as part of v8), the `createdBy` field was added to the `RecurringJob` class. Unfortunately, during the implementation of the serializer, it was overlooked that this field does not have a default and is therefore nullable.

With this, one of my fears regarding the ad-hoc approach of the Kotlin code to Java became true: nullability concerns. Ideally, we should verify that all non-null assumptions are correct before opening the flood gates. On the other hand, compatibility with other JobMappers is explicitly not guaranteed in the release notes. I tried running this fix against my v7.5.1 state and haven't found any other incompatibilities.


rdehuyss:
Hi @SIMULATAN - is there a way we can create a test for this you think? 

If not for all null cases, already for this case?

rdehuyss:
Hi @SIMULATAN - I think I may have found a better approach. The createdBy should not be null. So we're defaulting to API (as it will be overwritten anyway if it is created by an annotation).

Anyway - nice catch! I created PR https://github.com/jobrunr/jobrunr/pull/1329. Let me know what you think.

SIMULATAN:
Sounds reasonable. Since #1329 includes this PR, I'm going to close this one and continue the discussion there.

PR #1328, opened 2025-07-04
Bugfix - allow late init variables in Kotlin (corner case)

rdehuyss:
This bugfix allows late init variables in Kotlin, which depending on compiler settings use either a synthetic getter method or field access with a late init check called `UninitializedPropertyAccessException`.

Issue was found in project https://github.com/jobrunr/example-spring-kotlin but I'm unable to write a failing unit test for this (timeboxed for 2 hours).


wgroeneveld (approved):
Since we paired on it, I'm approving. 
Could in the future make it into a test by embedding all framework support subprojects into their own testbed projects.

PR #1329, opened 2025-07-04
Bugfix for createdBy being null when coming from JobRunr v7

rdehuyss:


core/src/testFixtures/resources/org/jobrunr/utils/mapper/existing-recurring-job-github-1327-input.json
SIMULATAN, line 1: Not sure how the tests are set up, but - unlike other jsons - this file isn't used in any test despite `testSerializeAndDeserializeRecurringJobsComingFrom5` referencing #1327 in the `@Because`. Is this on purpose?

rdehuyss: Leftover. I first created another test but as the contents was the same, I edited an existing test. Thanks for the review, I'll delete the file.

SIMULATAN (approved):
LGTM, thanks for the swift fix!

wgroeneveld (approved):
LGTM!

PR #1339, opened 2025-07-08
[NO-TASK] Performs deletion in smaller batches.

xethorn:
When a high number of jobs are scheduled within a short timeframe, there's typically a corresponding spike, in this deletion operation, of the number of rows to remove. 

The database resources can be overwhelmed and leave the transaction open for longer than expecting, creating a chain of performance issues. 

To improve the general performance, I updated the code to create smaller batches that are committed more frequently.

> [!NOTE]
> At the moment, this is a draft. I'm mostly looking for guidance on what you're expecting. For example: should the batch be something consumers can customize? Should


rdehuyss:
@cla-bot check

xethorn:
@rdehuyss 👋 how would you like me to proceed here?

PR #1340, opened 2025-07-08
sync schedule.durationBetweenSchedules() impl with Pro

wgroeneveld:
The validation in `durationBetweenSchedules` in `Schedule` is absent in OSS
This PR syncs its implementation with Pro


rdehuyss (approved):
LGTM!

PR #1351, opened 2025-07-15
Integrate Jest in the build

brainbaking:
The JobRunr repository has one JS test that does not run in the CI. It also has no ability to write unit tests for React. In this PR, we will address both issues by relying on:

- SWC for speedy compilation of tests that use ESM instead of CommonJS and JSX, both not easily supported by Node. This allows tests to be run from within an IDE without having to do `NODE_OPTIONS=--experimental-vm-modules` and without having to transpile using babel to fix the JSX problem. see https://swc.rs/docs/usage/jest
- Test Library for React assertions: see https://testing-library.com/docs/queries/bytext/

The `npmRunTest` Gradle task integrates `npm run test` that runs `jest` inside the JVM/CI build chain.

### DIFF
--- a/BuildDockerfile
+++ b/BuildDockerfile
@@ -2,20 +2,21 @@
 FROM maven:3.9.4-eclipse-temurin-21 AS maven-dist
 # build stage for Gradle
 FROM gradle:8.14.2-jdk21 AS gradle-dist
-
 # Main build stage
 FROM eclipse-temurin:21-jdk@sha256:88214b12ef97dcb4d44a96f23043041a78ab08fc035740309f1f0b026ce79940
+
+ARG NODE_VERSION=23.11.0
 
 RUN apt-get update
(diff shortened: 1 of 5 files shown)


brainbaking:
This also adds another step to the CI for basic ASCII JUnit reporting. I made a custom Woodpecker-CI plugin for this, see https://github.com/wgroeneveld/woodpecker-ascii-junit/ that picks up any JUnit XML and displays:

```
| Passed | Failed | Errored | Skipped | Total |
_______________________________________________
| 20     | 5      | 0       | 1       | 1     | 

Total time: 1.378s
```

Including error messages in case tests fail.

Jest tests also generate XML using the [jest-junit xml reporter](https://github.com/jest-community/awesome-jest/blob/main/README.md#reporters). For this to work I had to upgrade the used Node version in the docker build image.

rdehuyss (approved):
LGTM!

PR #1353, opened 2025-07-18
Improve info/error logging related to critical JobRunr lifecycle

wgroeneveld:
When running multiple background job servers, logging is very confusing if something goes wrong (e.g. a server stops/starts/restarts due to a ConcurrentModificationException, ...) because of missing or inconsistent name/ids. 

This tiny 🤏 PR adds a bit of verbosity to the important logger outputs making it easier for users and us to debug what's happening.


rdehuyss (approved):
LGTM!

PR #1367, opened 2025-08-06
Make groovy-json compile only

auloin:


auloin:
> Did you try to publish to maven central with this setting without releasing? This is the only thing I'm worried about.

Just did! All worked (sonatype refused it because I tested with a SNAPSHOT tag).

PR #1375, opened 2025-08-26
Improve performance by only fetching performance stats at most once every 10 ms

kabutz:
Here is a proposed fix (please check that it works correctly and is faster), where we utilize the CHM inside the JobServerStats to only query the OS MBean once every 10 ms.

### DIFF
--- a/core/src/main/java/org/jobrunr/server/jmx/JobServerStats.java
+++ b/core/src/main/java/org/jobrunr/server/jmx/JobServerStats.java
@@ -1,6 +1,7 @@
 package org.jobrunr.server.jmx;
 
 import javax.management.MBeanServer;
+import javax.management.ObjectName;
 import java.lang.management.ManagementFactory;
 import java.lang.management.OperatingSystemMXBean;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


rsobies:
@kabutz 
I still see cpu usage quite hight, like there is no difference
image

it looks like `isNotExpired`  always return false, I have put a log inside `if` and the log is never printed

kabutz:
I haven't been able to reproduce the issue on my side, except in a small benchmark. Could you perhaps add a counter like a LongAdder that counts how many times this method is called per second? Maybe the expiry time has to increase, although 10ms seems long enough to me.

rsobies:
@kabutz 
the intervals betwean invoking `getMXBeanValueAsDouble` and `getMXBeanValueAsLong` are much bigger then 10ms thats why cache is always expired, when i set the `EXPIRY_TIME_IN_NANOS` == 50sec (just for example) , then cache started to work

rsobies:
this line `final Object attribute = platformMBeanServer.getAttribute(operatingSystemMXBean.getObjectName(), name);`
is very expensive even it is called once per couple seconds

rsobies:
@kabutz 
you can reproduce it by using sample app from my github [https://github.com/rsobies/jobruntest2](https://github.com/rsobies/jobruntest2) just follow the steps in the readme

auloin:
@cla-bot check

rdehuyss:
Closing / cleanup stale PR's

PR #1386, opened 2025-09-17
Update dependencies

### LINKED ISSUE #1384: Update all dependencies
Update any dependency as long as there are no breaking change (i.e., to there latest minor version). Maven dependencies + npm dependencies.

auloin:
Closes #1384


core/src/main/resources/org/jobrunr/dashboard/frontend/package.json
```diff
@@ -7,16 +7,16 @@
...
-    "@mui/icons-material": "^5.17.1",
-    "@mui/material": "^5.17.1",
-    "apexcharts": "^4.7.0",
-    "cronstrue": "^2.57.0",
+    "@mui/icons-material": "^5.18.0",
+    "@mui/material": "^5.18.0",
+    "apexcharts": "^5.3.3",
+    "cronstrue": "^3.3.0",
     "mdi-material-ui": "^7.9.4",
     "react": "^18.3.1",
```
brainbaking, line 15: So we're not updating this one just yet I presume? Looks like `19.1` exists
Edit: in another PR of course. Ignore this.

framework-support/jobrunr-micronaut-feature/jobrunr-micronaut-annotations/build.gradle
```diff
@@ -8,7 +8,7 @@ plugins {
 }
 
 micronaut {
-    version '4.8.3'
+    version '4.9.3'
```
brainbaking, line 11: Does updating these and the Spring/Quarkus one mean our users have to update as well; how do we usually do this?

wgroeneveld (approved):

core/src/main/resources/org/jobrunr/dashboard/frontend/package.json
auloin: The upgrade is covered by #1388

framework-support/jobrunr-micronaut-feature/jobrunr-micronaut-annotations/build.gradle
auloin: Good question; from experience no (playing with the example projects where we don't always change the framework version). Unless we're using an API that's only in the later version, they won't need to upgrade. For Spring we explicitly set the framework dependency as `provided`.

PR #1389, opened 2025-09-19
Error-prone compile-type analysis

### LINKED ISSUE #1387: Add ErrorProne to increase code quality
https://errorprone.info/

brainbaking:
Closes #1387
This PR introduces https://errorprone.info/ to this repository

See the root `build.gradle` file. I've ignored some of the errors that are either not relevant for JobRunr or we cannot make use of because of JDK8. There are still a lot of warnings to gradually tackle.

Error-prone warnings still to solve together in future sessions:
- https://errorprone.info/bugpattern/TypeParameterUnusedInFormals we use `<T> T` a lot misusing generics to avoid casting
- https://errorprone.info/bugpattern/WaitNotInLoop for `await()`s in a lock that are not looped
- https://errorprone.info/bugpattern/NonApiType for `ArrayList` in the `Problem` types going to the frontend that rely on these specific implementations so changing to `List` will break the UI

Also includes extra rules by https://error-prone.picnic.tech/ that are easy to enable/disable; a subset is already prepared in the build file. Errorprone is disabled now to be able to gradually work on the above.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -104,17 +104,6 @@ configure(subprojects.findAll { !['platform'].contains(it.name) }) {
             disable("PatternMatchingInstanceof")                // we can't update to JDK9 just yet
             disable("StringCaseLocaleUsage")                    // we are never going to run into this issue
             disable("StringSplitter")                           // should not happen because of validation
-
-            // Picnic rules
-            disable("TimeZoneUsage")                            // expects deriving from Spring's Clock
-            disable("StaticImport")                             // clashes with error-prone's idea of sometimes not static importing
-
-            def picnicRefasterRulesToDisable = [
-                    "StringRules\$StringIsNullOrEmpty",         // assumes Guava
-                    "PreconditionsRules\$",                     // assumes Guava
-                    "NullRules\$RequireNonNullElse"             // we can't update to JDK9 just yet
-            ].collect { "(?!$it)" }.join("") + ".*"
-            option("-XepOpt:Refaster:NamePattern=^${picnicRefasterRulesToDisable}")
         }
     }
 
@@ -138,8 +127,6 @@ configure(subprojects.findAll { !['platform'].contains(it.name) }) {
         compileOnly 'org.apache.groovy:groovy-json:4.0.27'
 
         errorprone("com.google.errorprone:error_prone_core:2.40.0")
-        errorprone("tech.picnic.error-prone-support:error-prone-contrib:0.24.0")
-        errorprone("tech.picnic.error-prone-support:refaster-runner:0.24.0")
 
         testImplementation 'org.junit.jupiter:junit-jupiter'
         testImplementation 'io.github.artsok:rerunner-jupiter'


core/src/main/java/org/jobrunr/storage/navigation/AmountRequest.java
```diff
@@ -45,9 +48,9 @@ public String asString() {
 
     @VisibleFor("testing")
     public List<OrderTerm> getAllOrderTerms(Set<String> allowedOrderTerms) {
-        if (StringUtils.isNullOrEmpty(order)) return emptyList();
-        final String[] sortOns = order.split(",");
         List<OrderTerm> result = new ArrayList<>();
+        if (StringUtils.isNullOrEmpty(order)) return result;
```
auloin: Should we throw an exception. I don't think we expect null here (needs confirmation).

core/src/main/java/org/jobrunr/utils/exceptions/Exceptions.java
```diff
@@ -24,7 +24,7 @@ public static String getStackTraceAsString(Throwable exception) {
     /**
      * A {@link Supplier} that can throw checked Exceptions.
      *
-     * @param <T>
+     * @param <T> the generic supplier type.
```
auloin: Can't the line be dropped?

core/src/main/java/org/jobrunr/dashboard/server/http/url/RequestUrl.java
```diff
@@ -67,7 +76,7 @@ public <T> T fromQueryParams(Class<T> clazz) {
     }
 
     private Map<String, List<String>> initQueryParams(String url) {
-        if (!url.contains("?")) return Collections.emptyMap();
+        if (!url.contains("?")) return new HashMap<>();
```
auloin: For consistent return types

```suggestion
        if (!url.contains("?")) return new LinkedHashMap<>();
```

auloin: An alternative to fixing this would be using `Collections.unmodifiableMap` on the second return. This will depend on how we want to approach these: do we want to follow a immutability first rule?

core/src/test/java/org/jobrunr/dashboard/FrontEndDevelopment.java
```diff
@@ -81,7 +83,7 @@ public static void main(String[] args) throws Exception {
         BackgroundJob.<TestService>schedule(CarbonAware.at(now().plus(24, HOURS), Duration.ofHours(4)), x -> x.doWork(28));
         BackgroundJob.<TestService>schedule(CarbonAware.at(now().plus(48, HOURS), Duration.ofHours(4)), x -> x.doWork(52));
         BackgroundJob.<TestService>schedule(CarbonAware.at(now().plus(3, DAYS), Duration.ofHours(4)), x -> x.doWork(72));
-        BackgroundJob.<TestService>schedule(CarbonAware.between(LocalDate.now().atTime(20, 0), LocalDate.now().atTime(22, 0)), x -> x.doWork(72));
+        BackgroundJob.<TestService>schedule(CarbonAware.between(LocalDate.now(ZoneId.systemDefault()).atTime(20, 0), LocalDate.now(ZoneId.systemDefault()).atTime(22, 0)), x -> x.doWork(72));
```
auloin: Is it worth adding a LocalDateUtils to avoid the repetition of `LocalDate.now(ZoneId.systemDefault())` or `LocalDateTime.now(ZoneId.systemDefault())`?

core/src/testFixtures/java/org/junit/jupiter/extension/ForAllSubclassesExtension.java
```diff
@@ -77,7 +77,7 @@ private static Class findClassWithForAllSubclassesAnnotation(Class clazz) {
         return findClassWithForAllSubclassesAnnotation(clazz.getSuperclass());
     }
 
-    private static Method findMethodWithAnnotation(Class clazz, Class<? extends Annotation> annotation) {
+    private static Method findMethodWithAnnotation(Class unusedClass, Class<? extends Annotation> annotation) {
```
auloin: Why isn't it dropped?

core/src/main/java/org/jobrunr/storage/navigation/AmountRequest.java
brainbaking: Yup, can't happen, see constructor :-)

core/src/main/java/org/jobrunr/dashboard/server/http/url/RequestUrl.java
brainbaking: Since it's part of the requestUrl that should not be changeable (and we don't alter) I presume immutability is the way to go here

core/src/test/java/org/jobrunr/dashboard/FrontEndDevelopment.java
brainbaking: At the moment I don't think so; it only appears 3 times and that's in test code. `ZoneId.systemDefault()` is used more extensively in date-specific tests like `InstantUtilsTest` but that's on every possible date class (thai buddist, japanese, ...) and in other tests like the bgjob ones it switches between offsetdatetime, localdatetime, and zoneddatetime. 
I think having utility methods for all of these would introduce even more mess (at this moment)

PR #1391, opened 2025-09-24
Upgrade frontend dependencies

auloin:


core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/jobs/job-view.js
```diff
@@ -150,16 +150,16 @@ const JobView = (props) => {
...
                             <Typography color="textPrimary">{job.id}</Typography>
                         </Breadcrumbs>
-                        <Box my={3} style={{marginBottom: 0}}>
-                            <Card style={{display: "flex"}}>
-                                <CardContent style={{width: "100%"}}>
-                                    <Grid container spacing={3} justifyContent="space-between">
-                                        <Grid item xs={6}>
+                        <Box my={3}>
+                            <Card sx={{display: "flex", borderBottomLeftRadius: 0, borderBottomRightRadius: 0}}>
+                                <CardContent sx={{width: "100%", padding: 0, "&:last-child": {padding: 0}}}>
```
brainbaking, line 155: Curious what the reason is for these specific style changes; did they alter the way radius/padding works?

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/jobs/jobs-table.js
```diff
@@ -64,83 +63,75 @@ const JobsTable = ({jobPage, jobState, isLoading}) => {
     const isInAwaitingViewAndJobIsCarbonAware = (job) => jobState === "AWAITING" && isCarbonAwaitingState(getJobMostRecentState(job));
     const isInScheduledViewAndJobIsCarbonAware = (job) => jobState === "SCHEDULED" && isCarbonAwaitingState(getJobPreviousState(job));
 
-    return (
-        <> {isLoading
-            ? <LoadingIndicator/>
```
brainbaking, line 69: So no more loading indicator needed? `isLoading` can also go (3rd param)?

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/jobs/job-view.js
auloin: Good that you question this. I made this change to integrate nicely with the code block but since then I moved it inside the card. We no longer need to touch the border style.

core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/jobs/jobs-table.js
auloin: I've reverted the `isLoading` was never useful because the parent component `JobsView` already shows the loading indicator. But I'll keep this as is for now.

PR #1395, opened 2025-10-02
fix: disable text select of time scale pain in processing job step

miettal:
under testing.

Current implementation, User cannot copy only log output of "processing job" step, beause it's include time scale text (0 seconds, 3 seconds... for each line).
This PR provide copying only log output by disabling text select of time scale text.

### DIFF
--- a/core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/jobs/states/processing-state.js
+++ b/core/src/main/resources/org/jobrunr/dashboard/frontend/src/components/jobs/states/processing-state.js
@@ -22,6 +22,7 @@ const Console = styled("div")(() => ({
         textAlign: 'right',
         color: '#3885b7',
         margin: '-0.1em 0',
+        userSelect: 'none',
     },
     '& > dl dd': {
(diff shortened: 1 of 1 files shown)


auloin:
@cla-bot check

auloin:
Hi @miettal, any update on this PR? I think we can move forward with this.

It may also be useful to offer a copy to clipboard feature with different options as some people may want to include the actual timestamp, something similar to how logback prints the logs:

```
16:06:09.031 INFO -- Entering application.
16:06:09.046 DEBUG -- Did it again!
16:06:09.046 INFO -- Exiting application.
```

(we could also allow a user to change the view using different timestamp styles)

miettal:
I didn't test well. but anyway, I change this PR state to ready.

I think "copy button" is also nice choices!

auloin:
> I didn't test well. but anyway, I change this PR state to ready.

Thanks, it looks good to me.

> I think "copy button" is also nice choices!

Ok, I'll note this down for potential improvement.

auloin (approved):

PR #1402, opened 2025-10-10
search jobs by states

hanouard:
## Motivation
Previously the API implicitly handled only one state. This feature enables richer dashboards and monitoring by allowing multi-state aggregation.


rdehuyss:
I'm sorry but as search is part of pro, we cannot accept this PR.

Typically contributions are first discussed in discussions to avoid these kind of misunderstandings.

PR #1405, opened 2025-10-20
Disable browser caching for `/api` endpoints

auloin:


core/src/main/java/org/jobrunr/dashboard/server/sse/SseExchange.java
```diff
@@ -17,7 +17,7 @@ public class SseExchange implements AutoCloseable {
     public SseExchange(HttpExchange httpExchange) throws IOException {
         this.writer = new BufferedWriter(new OutputStreamWriter(httpExchange.getResponseBody(), StandardCharsets.UTF_8));
         Headers responseHeaders = httpExchange.getResponseHeaders();
-        responseHeaders.add("Cache-Control", "no-cache,public");
+        responseHeaders.add("Cache-Control", "no-cache, no-store, private");
```
brainbaking, line 20: Does this mean we should also add/modify the appropriate headers for Spring in `JobRunrSpringSseController`? From what I understand the controller method returning an `SseEmitter` should also set the correct response header?

auloin: Those classes are not available in this repo but yes indeed we also need to add those to the embedded controllers in PRO.

PR #1406, opened 2025-10-20
Add CSP headers

auloin:


auloin:
> Not sure about this but what about the static file resource handlers for our compatible frameworks? They don't utilize `handle()` so I think we should also add that response header there

Those classes are not available in this repo but yes indeed we also need to add those to the embedded controllers in PRO.

gikaragia:
Hello @auloin @brainbaking, after upgrading to 8.3.1 from 7.5.x we can't access the dashboard anymore. All requests to retrieve static files are blocked:

Screenshot 2026-01-13 at 4 46 07 PM

It seems to me that this is a bug on Jobrunr, correct? Or is there a misconfiguration on our end? If it matters for prioritization, we own a Pro license.

auloin:
Hi @gikaragia, please send a mail to support@jobrunr.io and add enough info so we can reproduce this issue (how is your setup like, etc.).

auloin:
@gikaragia we haven't heard back from you. Do you still have this issue? It's really strange and we'd like to find out why it's happening.

gikaragia:
Hey @auloin, yes, the issue generally exists, but we have applied a custom fix to our installation to not block the upgrade. I have communicated the details as you requested by email to support@jobrunr.io

auloin:
@gikaragia something must have went wrong because we still can't find the email in our inbox. Do you mind trying again? Just in case, could you also CC developers@jobrunr.io?

gikaragia:
Sure @auloin! Forwarded it to developers@jobrunr.io

PR #1415, opened 2025-10-29
First support for Spring Boot 4

rdehuyss:


framework-support/jobrunr-spring-boot-4-starter/build.gradle
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                }
+
+                withXml {
+                    def pomNode = asNode()
+                    pomNode.remove(pomNode.dependencyManagement)
+                    pomNode.dependencies.'*'.each() {
+                        if (it.optional.text() == 'true') {
+                            it.parent().remove(it)
+                        } else if (it.artifactId.text() == 'spring-boot-starter') {
```
auloin: Should we change to the bellow? I also wonder what will happen if it's not provided since it's not required anymore.
```suggestion
                        } else if (it.artifactId.text() == 'spring-boot-starter-classic') {
```

platform/build.gradle
```diff
@@ -57,7 +59,7 @@ dependencies {
         api 'org.testcontainers:testcontainers-db2:2.0.1'
         api 'org.testcontainers:testcontainers-mongodb:2.0.1'
         api 'com.microsoft.playwright:playwright:1.56.0'
-        api 'org.wiremock:wiremock:3.13.1'
+        api 'org.wiremock:wiremock-standalone:3.13.1'
```
auloin: @wgroeneveld why the switch to standalone? from the docs it seems to be designed to run as a separate process, which doesn't seem to be how we're using it.

core/src/main/java17/org/jobrunr/utils/mapper/jackson3/Jackson3JsonMapper.java
```diff
@@ -0,0 +1,67 @@
...
+    private final tools.jackson.databind.json.JsonMapper jsonMapper;
+
+    public Jackson3JsonMapper() {
+        var typeValidator = BasicPolymorphicTypeValidator.builder()
+                .allowIfSubType("org.jobrunr.")
+                .allowIfSubType("java.util.concurrent.")
+                .allowIfSubTypeIsArray()
+                .build();
+
+        this.jsonMapper = tools.jackson.databind.json.JsonMapper.builder()
```
auloin: Should we make this extendable by allowing the end user to pass a builder? Could their config conflict with our? For instance in case they change getters visibility...

framework-support/jobrunr-spring-boot-4-starter/build.gradle
auloin: I think this is something for @rdehuyss

platform/build.gradle
brainbaking: This is important as there are wiremock dependencies in `src/testFixtures` as I moved stuff to reuse it. Without standalone, the embedded Jetty container isn't there that's implicitly included because of other deps we have.

core/src/main/java17/org/jobrunr/utils/mapper/jackson3/Jackson3JsonMapper.java
brainbaking: Like we do in the Jackson 2 constructor? I don't know; I'd rather not if it's not needed but I guess it can be done

Things like `.enable()` and `.changeDefaultVisibility()` that we call should override their settings (`activateDefaultTypingAsProperty()` looks to be a setter behind the scenes as well) so that's good, I don't know what happens when they have their own mixins and such. 

Should we do it?

platform/build.gradle
auloin: I see... I was able to solve the build error with:

```
testImplementation 'org.wiremock:wiremock'
// ...
testFixturesCompileOnly 'org.wiremock:wiremock'
```

What do you think of this approach?

brainbaking: Fine by me!

PR #1429, opened 2025-11-12
Add JSpecify annotations on public methods

timtebeek:
- CLA https://github.com/jobrunr/clabot-config/pull/65

As discussed last week: maybe some nullability annotations could help clarify where elements might be null.

- Public methods that `return null`, or return the response of a method known to be nullable, like `Map.get(K)`, now get a `@Nullable` added to their return type
- Public methods with arguments that are null checked in the method body, see `@Nullable` added to the corresponding parameters
- A handful of JetBrains nullable annotations are replaced with JSpecify

Based on an automation: https://docs.openrewrite.org/recipes/java/jspecify/jspecifybestpractices

### DIFF
--- a/core/src/main/java/org/jobrunr/scheduling/JobScheduler.java
+++ b/core/src/main/java/org/jobrunr/scheduling/JobScheduler.java
@@ -14,7 +14,7 @@ import org.jobrunr.jobs.lambdas.JobLambdaFromStream;
 import org.jobrunr.scheduling.carbonaware.CarbonAwarePeriod;
 import org.jobrunr.scheduling.interval.Interval;
 import org.jobrunr.storage.StorageProvider;
-import org.jspecify.annotations.Nullable;
+import org.jspecify.annotations.NonNull;
 
 import java.time.Duration;
 import java.time.Instant;
@@ -64,10 +64,8 @@ public class JobScheduler extends AbstractJobScheduler {
         this(storageProvider, new JobDetailsAsmGenerator(), jobFilters);
     }
 
-    public JobScheduler(StorageProvider storageProvider, @Nullable JobDetailsGenerator jobDetailsGenerator, List<JobFilter> jobFilters) {
+    public JobScheduler(StorageProvider storageProvider, @NonNull JobDetailsGenerator jobDetailsGenerator, List<JobFilter> jobFilters) {
         super(storageProvider, jobFilters);
-        if (jobDetailsGenerator == null)
-            throw new IllegalArgumentException("A JobDetailsGenerator is required to use the JobScheduler.");
         this.jobDetailsGenerator = jobDetailsGenerator;
         BackgroundJob.setJobScheduler(this);
     }
--- a/core/src/main/java/org/jobrunr/scheduling/carbonaware/CarbonAwareScheduleMargin.java
+++ b/core/src/main/java/org/jobrunr/scheduling/carbonaware/CarbonAwareScheduleMargin.java
@@ -1,5 +1,6 @@
 package org.jobrunr.scheduling.carbonaware;
 
+import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 
 import java.time.Duration;
@@ -18,10 +19,7 @@ public class CarbonAwareScheduleMargin {
     private final Duration marginBefore;
     private final Duration marginAfter;
 
-    public CarbonAwareScheduleMargin(@Nullable Duration marginBefore, @Nullable Duration marginAfter) {
-        if (marginBefore == null || marginAfter == null) {
-            throw new IllegalArgumentException(format("Expected marginBefore (='%s') and marginAfter (='%s') to be non-null.", marginBefore, marginAfter));
-        }
+    public CarbonAwareScheduleMargin(@NonNull Duration marginBefore, @NonNull Duration marginAfter) {
         if (marginBefore.isNegative() || marginAfter.isNegative()) {
             throw new IllegalArgumentException(format("Expected marginBefore (='%s') and marginAfter (='%s') to be positive Durations.", marginBefore, marginAfter));
         }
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/jobrunr/dashboard/server/http/url/RequestUrl.java
```diff
@@ -38,7 +39,7 @@ public Map<String, String> getParams() {
         return params;
     }
 
-    public String param(String paramName) {
+    public @Nullable String param(String paramName) {
         return params.get(paramName);
```
timtebeek, line 43: This is an example of where calling `.get(...)` on a `Map<String, String> params` results in the return type being nullable, as there's no hard guarantee that any `paramName` passed in would be present in the Map. If you know from context that is always the case then the annotation could be dropped.

core/src/main/java/org/jobrunr/jobs/details/instructions/BAStoreOperandInstruction.java
```diff
@@ -14,7 +15,7 @@ public void load() {
     }
 
     @Override
-    public Object invokeInstruction() {
+    public @Nullable Object invokeInstruction() {
         return null;
```
timtebeek, line 19: At least some of the implementations here use `return null`; would it make sense to pull up the annotation to the super class or interface for this method?

core/src/main/java/org/jobrunr/scheduling/JobScheduler.java
```diff
@@ -63,7 +64,7 @@ public JobScheduler(StorageProvider storageProvider, List<JobFilter> jobFilters)
         this(storageProvider, new JobDetailsAsmGenerator(), jobFilters);
     }
 
-    public JobScheduler(StorageProvider storageProvider, JobDetailsGenerator jobDetailsGenerator, List<JobFilter> jobFilters) {
+    public JobScheduler(StorageProvider storageProvider, @Nullable JobDetailsGenerator jobDetailsGenerator, List<JobFilter> jobFilters) {
         super(storageProvider, jobFilters);
         if (jobDetailsGenerator == null)
             throw new IllegalArgumentException("A JobDetailsGenerator is required to use the JobScheduler.");
```
timtebeek: This flow makes me doubt if we should even annotate the argument as nullable; what do you think about removing the `@Nullable` and this check + throw?

core/src/main/java/org/jobrunr/scheduling/carbonaware/CarbonAwareScheduleMargin.java
```diff
@@ -16,7 +18,7 @@ public class CarbonAwareScheduleMargin {
     private final Duration marginBefore;
     private final Duration marginAfter;
 
-    public CarbonAwareScheduleMargin(Duration marginBefore, Duration marginAfter) {
+    public CarbonAwareScheduleMargin(@Nullable Duration marginBefore, @Nullable Duration marginAfter) {
         if (marginBefore == null || marginAfter == null) {
             throw new IllegalArgumentException(format("Expected marginBefore (='%s') and marginAfter (='%s') to be non-null.", marginBefore, marginAfter));
         }
```
timtebeek: Same as above; it's odd seeing nullable + check + throws; What would you think of using `@NonNull` here instead? One can even add `@NullMarked` on an entire package as a way to avoid adding `@NonNull`.

timtebeek:
I've left some self review comments that will apply beyond just the cases pointed out, but figured get your input before continuing. I think it's done a decent job of discovering patterns where explicit null-ability could help.

rdehuyss:
@cla-bot check

core/src/main/java/org/jobrunr/scheduling/JobScheduler.java
```diff
@@ -63,10 +64,8 @@ public JobScheduler(StorageProvider storageProvider, List<JobFilter> jobFilters)
         this(storageProvider, new JobDetailsAsmGenerator(), jobFilters);
     }
 
-    public JobScheduler(StorageProvider storageProvider, JobDetailsGenerator jobDetailsGenerator, List<JobFilter> jobFilters) {
+    public JobScheduler(StorageProvider storageProvider, @NonNull JobDetailsGenerator jobDetailsGenerator, List<JobFilter> jobFilters) {
         super(storageProvider, jobFilters);
-        if (jobDetailsGenerator == null)
-            throw new IllegalArgumentException("A JobDetailsGenerator is required to use the JobScheduler.");
```
timtebeek, line 69: We now remove these types of checks after adding the `@NonNull` annotation, since they should not happen. The confusing alternative would be to annotate as `@Nullable`, and then throw whenever it is null. 🙃

core/src/main/java/org/jobrunr/scheduling/carbonaware/CarbonAwareScheduleMargin.java
```diff
@@ -16,18 +19,15 @@ public class CarbonAwareScheduleMargin {
     private final Duration marginBefore;
     private final Duration marginAfter;
 
-    public CarbonAwareScheduleMargin(Duration marginBefore, Duration marginAfter) {
-        if (marginBefore == null || marginAfter == null) {
-            throw new IllegalArgumentException(format("Expected marginBefore (='%s') and marginAfter (='%s') to be non-null.", marginBefore, marginAfter));
-        }
+    public CarbonAwareScheduleMargin(@NonNull Duration marginBefore, @NonNull Duration marginAfter) {
```
timtebeek, line 22: Optionally there's also the `@NullMarked` annotation which you can add to the entire package, meaning you only have to annotate nullable parameters, and can leave out these `@NonNull` annotations.

brainbaking: Hi Tim, as far as I understood checkers like NullAway that read JSpecify annotations, aren't they non null by default? Then null marked isn't needed?

timtebeek: If that's universal beyond NullAway then indeed we would not have to be explicit here. Just not sure if those defaults apply everywhere, or only when the package is annotated. I'm my case I have not used NullAway, but mostly Intellij IDEA.

rdehuyss:
Hi @timtebeek - could we have push access to your branch?

timtebeek:
> Hi @timtebeek - could we have push access to your branch?

Should be enabled! I tend to check out PRs via Intellij, which makes that easy

timtebeek:
As discussed I'll close this effort; it can always be recreated quickly if desired; Mostly we'd need to work out whether gradual adoption is ok, by for instance first and only annotating public methods that explicitly `return null`.

PR #1454, opened 2025-12-07
Add defensive null check in setMethodName() for Kotlin lambda bytecode edge case

### LINKED ISSUE #1453: [BUG] NPE in KotlinJobDetailsBuilder when parsing Kotlin lambda compiled with Bazel rules_kotlin
### JobRunr Version

8.3.0 (with jobrunr-kotlin-2.1-support 8.3.0)

### JDK Version

OpenJDK 17.0.11 (Temurin)

### Your SQL / NoSQL database

H2 in-memory (but the bug occurs before any database interaction)

### What happened?

When enqueueing a Kotlin lambda in a project compiled with Bazel's `rules_kotlin`, JobRunr throws an NPE in `KotlinJobDetailsBuilder`. The identical code compiled with Gradle works correctly.

**Stack trace:**
```
java.lang.NullPointerException: Cannot invoke "String.endsWith(String)" because "name" is null
    at org.jobrunr.jobs.details.JobDetailsBuilder.setMethodName(JobDetailsBuilder.java:114)
    at org.jobrunr.jobs.details.JobDetailsBuilder.<init>(JobDetailsBuilder.java:41)
    at org.jobrunr.jobs.details.JobDetailsBuilder.<init>(JobDetailsBuilder.java:31)
    at org.jobrunr.jobs.details.KotlinJobDetailsBuilder.<init>(KotlinJobDetailsBuilder.java:13)
    at org.jobrunr.jobs.details.KotlinJobDetailsFinder.<init>(KotlinJobDetailsFinder.java:38)
    at org.jobrunr.jobs.details.JobDetailsAsmGenerator.toJobDetails(JobDetailsAsmGenerator.java:20)
```

All three Kotlin lambda forms trigger the NPE:
- `jobScheduler.enqueue { myService.doSomething() }`
- `jobScheduler.enqueue<MyService> { it.doSomething() }`
- `jobScheduler.enqueue(myService::doSomething)`

**Expected behavior:**
JobRunr should either:
1. Successfully parse the lambda bytecode from rules_kotlin, or
2. Fail with a clear error message indicating unsupported bytecode pattern (not an NPE)

The bytecode produced by rules_kotlin is valid - the application runs correctly. However, `KotlinJobDetailsFinder`'s ASM-based parsing fails to extract the method name from the lambda.

### How to reproduce?

**Reproduction repository:** https://github.com/johnnymo87/jobrunr-kotlin-bazel-repro

The repo contains two builds of identical Kotlin code:
- `gradle-project/` - Gradle build (works)
- Root directory - Bazel build with rules_kotlin 2.1.0 (fails)

**To reproduce:**
```bash
git clone https://github.com/johnnymo87/jobrunr-kotlin-bazel-repro.git
cd jobrunr-kotlin-bazel-repro

# Gradle - works
cd gradle-project
./gradlew test
# Result: PASSES

# Bazel - fails (from repo root)
cd ..
bazel test //:JobrunrKotlinNpeReproTest
# Result: NPE in KotlinJobDetailsFinder
```

**Minimal code that fails:**
```kotlin
@Service
class MyService {
    fun doSomething(): String = "done"
}

// In a test:
@SpringBootTest
class JobrunrKotlinNpeReproTest {
    @Autowired lateinit var jobScheduler: JobScheduler
    @Autowired lateinit var myService: MyService

    @Test
    fun `enqueue kotlin lambda`() {
        jobScheduler.enqueue { myService.doSomething() }  // NPE here
    }
}
```

**Environment:**
- Kotlin 2.1.0
- Spring Boot 3.4.2
- Bazel 7.4.1 + rules_kotlin 2.1.0

**What I've tried:**
1. Setting `java_parameters = True` in rules_kotlin - still fails
2. Matching `jvm_target = "17"` - still fails
3. Various `kotlinc_opts` combinations - still fails

The only difference is the build tool (Gradle vs Bazel/rules_kotlin).

**Workarounds found:**
1. JobRequest/JobRequestHandler API - bypasses bytecode analysis entirely
2. Java bridge - move the lambda to Java code so `JavaJobDetailsFinder` is used instead

### Relevant log output

```shell

```

johnnymo87:
This PR adds a null check to `JobDetailsBuilder.setMethodName()` that throws a helpful `IllegalStateException` instead of an opaque `NullPointerException`.

### Context

When Kotlin lambdas are compiled with `-Xlambdas=class` (which generates anonymous inner classes) rather than `-Xlambdas=indy` (the Kotlin 2.x default using `invokedynamic`), the bytecode structure differs enough that `KotlinJobDetailsFinder` can't extract the method name. This results in `null` reaching `setMethodName()`, causing:

```
java.lang.NullPointerException: Cannot invoke "String.endsWith(String)" because "name" is null
```

The root cause is in Bazel's `rules_kotlin`, which defaults to `x_lambdas = "class"`. I've opened [bazelbuild/rules_kotlin#1419](https://github.com/bazelbuild/rules_kotlin/pull/1419) to document the workaround there.

### What this PR does

Adds a defensive null check that throws an actionable error message pointing users toward the compiler flag mismatch. The message mentions `-Xlambdas=class` vs `-Xlambdas=indy` specifically so users can search for and apply the fix.

### What this PR doesn't do

This doesn't add support for parsing `-Xlambdas=class` bytecode—that would require more significant changes to `KotlinJobDetailsFinder`. This is a small, low-risk improvement to developer experience.

### Why `IllegalStateException`?

I followed the existing pattern in this method, which throws `IllegalArgumentException` for the `$default` suffix check. `IllegalStateException` felt appropriate here since we're detecting an invalid state (missing method name) rather than an invalid argument value. Happy to change this if you prefer a different exception type.

---

Fixes #1453

I'm happy to iterate on the approach if you have feedback or would prefer a different direction.

### DIFF
--- a/core/src/main/java/org/jobrunr/jobs/details/JobDetailsBuilder.java
+++ b/core/src/main/java/org/jobrunr/jobs/details/JobDetailsBuilder.java
@@ -110,6 +110,14 @@ public abstract class JobDetailsBuilder {
     }
 
     public void setMethodName(String name) {
+        if (name == null) {
+            throw new IllegalStateException(
+                "Could not determine method name from lambda bytecode. " +
+                "This may occur when Kotlin lambdas are compiled with '-Xlambdas=class' instead of " +
+                "the default '-Xlambdas=indy'. Please ensure your build configuration uses the Kotlin " +
+                "compiler's default lambda compilation mode."
+            );
+        }
         if (name.endsWith("$default") && classExists("kotlin.KotlinVersion")) {
             throw new IllegalArgumentException("Unsupported lambda", new UnsupportedOperationException("You are (probably) using Kotlin default parameter values which is not supported by JobRunr."));
         }
--- a/core/src/test/java/org/jobrunr/jobs/details/JobDetailsBuilderTest.java
+++ b/core/src/test/java/org/jobrunr/jobs/details/JobDetailsBuilderTest.java
@@ -50,6 +50,16 @@ class JobDetailsBuilderTest {
                 .hasMessage("The lambda you provided is not valid.");
     }
 
+    @Test
+    void setMethodNameWithNullThrowsHelpfulException() {
+        final JobDetailsBuilder jobDetailsBuilder = getJobDetailsBuilder();
+
+        assertThatThrownBy(() -> jobDetailsBuilder.setMethodName(null))
+                .isInstanceOf(IllegalStateException.class)
+                .hasMessageContaining("Could not determine method name from lambda bytecode")
+                .hasMessageContaining("-Xlambdas=");
+    }
+
     private JobDetailsBuilder getJobDetailsBuilder() {
         return new JobDetailsBuilder(Arrays.asList("World", null), toFQClassName("org/jobrunr/examples/webapp/api/JobController"), "lambda$simpleJob$4ffb5ff$1") {
 


rdehuyss:
@cla-bot check

rdehuyss:
Superseded by #1460

PR #1459, opened 2025-12-18
Fix `assertJobExists` to keep job parameter not deserializable exception

auloin:
This PR contains two changes
1. When a job parameter is not deserializable, `assertJobExists` used to throw `JobMethodNotFoundException` with a helpful `jobParameterNotDeserializableStackTrace`. The message was not set in recent versions. The PR fixes that and adds an extra test in `BackgroundJobByJobLambdaTest` to prevent it from happening again.
2. When a job parameter is not deserializable, we keep the generic `ObjectNode` to be able to recover. The previous behaviour also changes the job signature to something like `my.Class.method(tools.jackson.databind.node.ObjectNode)`. This breaks application relying on the job signature to find jobs. The new approach is not changing the JobSignature.


core/src/main/java/org/jobrunr/utils/JobUtils.java
```diff
@@ -60,7 +60,14 @@ public static Method getJobMethod(Class<?> jobClass, JobDetails jobDetails) {
...
+        String methodName = jobDetails.getMethodName();
+        String[] jobParameterTypesNames = getDeserializableJobParametersForSignature(jobDetails);
+        if (jobParameterTypesNames.length == jobDetails.getJobParameters().size()) {
+            assertJobExists(className, methodName, jobParameterTypesNames);
+        } else {
+            throw new JobMethodNotFoundException(jobDetails);
+        }
     }
 
     public static void assertJobExists(String jobSignature) {
```
auloin, line 73: This method (`assertJobExists(String jobSignature)`) was added recently and is no longer used internally other than in tests other than in tests. Should we remove it?

PR #1461, opened 2025-12-19
Auto configure KotlinxSerializationJsonMapper

auloin:


rdehuyss:
I don't understand why the original code is not working.

Spring should take into account the 
`@ConditionalOnMissingClass(value = {"kotlinx.serialization.json.Json"})` and `@ConditionalOnClass(value = {kotlinx.serialization.json.Json.class, KotlinxSerializationJsonMapper.class})` to bootstrap the relevant JsonMapper?

rdehuyss:
Also fixes #1463.

PR #1478, opened 2026-01-30
Fix JobRunrRecurringJobRecorder

auloin:
Fixes Quarkus `@Recorder` constructor injection misuse (the issue was reported to us via email).

The fix doesn't require upgrading the Quarkus version. ~I've nonetheless bumped the version in `quarkus-jobrunr-tests` to keep our e2e tests green against latest Quarkus.~


cjgratacos:
Thank you for this one, this is preventing clean update to newer version of Quarkus for applications that are using Jobrunr.

rdehuyss (approved):
LGTM! 

RuntimeValue already exists for 7 years so I don't think we will break people on older versions.

Issue #1480, opened 2026-02-04
[Feature] `DatabaseCreator` should log statements with table prefixes in case of an exception


auloin:
**Is your feature request related to a problem? Please describe.**
When a migration fails, the error log at `DatabaseCreator.java:221` outputs the original SQL statement (without table prefix) instead of the actual executed statement (with table prefix applied):

```java
LOGGER.warn("Error running statement: {}", statement, e);  // should be updatedStatement
```

This is confusing because users see `CREATE TABLE jobrunr_backgroundjobservers_new` in the log, but the actual table in the database is `YBD_jobrunr_backgroundjobservers_new` (with prefix).

**Describe the solution you'd like**
The statement that contains the table prefix should be logged instead.

**Describe alternatives you've considered**
NA

**Additional context**
Requested by @tan9

Dan-DH:
Added in commit 2a25097

PR #1481, opened 2026-02-04
Log statement with table prefix in case of an exception in `runMigrationStatement`

Dan-DH:
For issue #1480


auloin:
@cla-bot check

PR #1489, opened 2026-02-14
Improve startup performance by batching migration checks

tan9:
## Summary

When `runMigrations()` is called and all migrations are already applied (the common steady-state case), the current implementation queries the `jobrunr_migrations` table individually for each migration file using `isMigrationApplied()`. With 17 migrations per database in OSS (and potentially more in Pro), this creates an N+1 query pattern that adds significant startup latency — especially noticeable on databases like Oracle where each round-trip has higher overhead.

This PR replaces the per-migration `SELECT count(*) ... WHERE script = ?` queries with a single `SELECT script FROM jobrunr_migrations` that loads all applied migrations into a per-invocation cache, then `isMigrationApplied()` reads from that cache.

### Changes

- **New `loadAppliedMigrations()`**: Loads all applied migration script names in one query into `appliedMigrationsCache`, also detects duplicate migrations (throwing `IllegalStateException`)
- **`runMigrations()`**: Calls `loadAppliedMigrations()` before filtering, then continues to use `isMigrationApplied()` in the stream filter — **preserving the Template Method pattern** so subclass overrides are respected
- **`isMigrationApplied()`**: Uses cached results when available; falls back to the original per-migration query when called standalone (backward compatible)
- When `isMigrationsTableMissing()` returns `true`, the bulk query is skipped entirely (uses `Collections.emptySet()` as cache)
- Cache is cleared after filtering to avoid stale data

### Impact

| Scenario | Before | After |
|----------|--------|-------|
| Steady state (all applied) | N+1 queries (17 individual SELECTs) | 1 single SELECT |
| Fresh install | Same (skipped via `isMigrationsTableMissing`) | Same |
| Oracle round-trips | ~68+ (connection + autocommit + query + close per migration) | ~2 |

All 19 existing tests pass without modification.

### DIFF
--- a/core/src/main/java/org/jobrunr/storage/sql/common/DatabaseCreator.java
+++ b/core/src/main/java/org/jobrunr/storage/sql/common/DatabaseCreator.java
@@ -61,7 +61,6 @@ public class DatabaseCreator {
     private final TablePrefixStatementUpdater tablePrefixStatementUpdater;
     private final DatabaseMigrationsProvider databaseMigrationsProvider;
     private final MigrationsTableLocker migrationsTableLocker;
-    private Set<String> appliedMigrationsCache;
 
     public static void main(String[] args) {
         if (args.length < 3) {
@@ -114,18 +113,15 @@ public class DatabaseCreator {
     }
 
     public void runMigrations() {
-        if (isMigrationsTableMissing()) {
-            appliedMigrationsCache = Collections.emptySet();
-        } else {
-            loadAppliedMigrations();
-        }
+        Set<String> appliedMigrations = isMigrationsTableMissing()
+                ? Collections.emptySet()
+                : loadAppliedMigrations();
         List<SqlMigration> migrationsToRun = getMigrations()
                 .filter(migration -> migration.getFileName().endsWith(".sql"))
                 .sorted(comparing(SqlMigration::getFileName))
-                .filter(migration -> !isMigrationApplied(migration))
+                .filter(migration -> !appliedMigrations.contains(migration.getFileName()))
                 .collect(toList());
         runMigrations(migrationsToRun);
-        loadAppliedMigrations();
     }
 
     public void validateTables() {
@@ -246,7 +242,7 @@ public class DatabaseCreator {
      * Loads all applied migration script names in a single query into cache,
      * replacing per-migration queries (N+1 problem).
      */
-    private void loadAppliedMigrations() {
+    protected Set<String> loadAppliedMigrations() {
         try (final Connection conn = getConnection();
              final Statement st = conn.createStatement();
              final ResultSet rs = st.executeQuery("select script from " + tablePrefixStatementUpdater.getFQTableName("jobrunr_migrations"))) {
@@ -260,7 +256,7 @@ public class DatabaseCreator {
                             "Please verify your migrations manually, cleanup the migrations_table and remove duplicate entries.");
                 }
             }
-            appliedMigrationsCache = migrationCounts.keySet();
+            return migrationCounts.keySet();
         } catch (SQLException sqlException) {
             LOGGER.debug("Error loading applied migrations", sqlException);
             throw new StorageException(sqlException);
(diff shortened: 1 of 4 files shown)


rdehuyss:
Hi @tan9! Thanks for this improvement!

core/src/main/java/org/jobrunr/storage/sql/common/DatabaseCreator.java
```diff
@@ -232,15 +242,39 @@ protected void updateMigrationsTable(Connection connection, SqlMigration migrati
...
         return migration.getFileName().endsWith("v000__create_migrations_table.sql");
     }
 
     protected boolean isMigrationApplied(SqlMigration migration) {
+        if (appliedMigrationsCache != null) {
+            return appliedMigrationsCache.contains(migration.getFileName());
+        }
         try (final Connection conn = getConnection();
              final PreparedStatement pSt = conn.prepareStatement("select count(*) from " + tablePrefixStatementUpdater.getFQTableName("jobrunr_migrations") + " where script = ?")) {
             boolean result = false;
```
rdehuyss: Do we still need the rest of this method? Can appliedMigrationsCache be null?

tan9: Good catch! Yes, `appliedMigrationsCache` can be `null` — it's set back to `null` at the end of `runMigrations()` (after the filtering is done). Since `isMigrationApplied` is `protected`, it could be called outside the `runMigrations()` flow (e.g., in tests like `DatabaseCreatorTest`), so the DB query fallback ensures it still works in that case.

That said, if you'd prefer a cleaner approach, I can remove the `appliedMigrationsCache = null` reset and drop the DB query fallback entirely — the `DatabaseCreator` is short-lived anyway, so keeping the cache around shouldn't be a concern. Would you prefer that simplification?

rdehuyss: Well, I wanted to get your thoughts on it...

We like your approach and we also like simple code 😉. Feel free to do so if you have the time, otherwise we'll go ahead and branch of your branch.

And, most of all, thanks for contributing back!

tan9: Done! I've simplified `isMigrationApplied` to a single cache lookup and removed the DB query fallback. The cache is now kept populated after `runMigrations()` completes instead of being nulled out.

tan9:
@rdehuyss Heads up — the previous push (050aa09) had a CI failure because `DefaultSqlStorageProviderTest` was missing a mock for `Statement.executeQuery()`, which our new `loadAppliedMigrations()` call at the end of `runMigrations()` now requires. The latest push (1d59472) fixes this by adding the missing mock setup (same fix as `DatabaseCreatorTablePrefixTest`). CI is green now.

core/src/main/java/org/jobrunr/storage/sql/common/DatabaseCreator.java
```diff
@@ -109,13 +114,18 @@ public DatabaseCreator(ConnectionProvider connectionProvider, String tablePrefix
...
+            loadAppliedMigrations();
+        }
         List<SqlMigration> migrationsToRun = getMigrations()
                 .filter(migration -> migration.getFileName().endsWith(".sql"))
                 .sorted(comparing(SqlMigration::getFileName))
-                .filter(x -> isMigrationTableMissing || isNewMigration(x))
+                .filter(migration -> !isMigrationApplied(migration))
                 .collect(toList());
         runMigrations(migrationsToRun);
+        loadAppliedMigrations();
```
auloin: Can we drop this? Tests will fail but we probably can also replace `isMigrationApplied` by `loadAppliedMigrations`. I believe `isMigrationApplied` is protected to ease testing, then if we offer `Set<String> loadAppliedMigrations` we could refactor the tests without much effort? The change would also allow us to do without the `appliedMigrationsCache` state.

Suggested changes:
1. We drop `isMigrationApplied` entirely (It's unlikely anyone has extended the `DatabaseCreator` or calling `isMigrationApplied`)
2. As alternative, we change `loadAppliedMigrations` to `protected Set<String> loadAppliedMigrations` that can be used in tests
3. Since we don't have `isMigrationApplied` there is no need for the state `appliedMigrationsCache`.

What do you think?

tan9: Done — removed `isMigrationApplied` and `appliedMigrationsCache`, `loadAppliedMigrations` now returns `Set<String>` directly. 95358d9a

Issue #1491, opened 2026-02-20
[BUG] ForkJoinPool with SecurityManager generate AccessControlException


keldkemp:
### JobRunr Version

8.4.2

### JDK Version

eclipse-temurin 17.0.17

### Your SQL / NoSQL database

PostgreSQL 13.3

### What happened?

Hello. In `org.jobrunr.storage.AbstractStorageProvider#notifyJobStatsOnChangeListeners`, `ForkJoinPool.commonPool()` is typically used.

We have a `SecurityManager` in our project (Yes, I know it's deprecated. But we can't drop it yet). By default, the `commonPool` uses a default thread factory. This means the threads have no permissions enabled, which leads to an error.

Example error:   
```java
java.util.concurrent.CompletionException: org.jobrunr.storage.StorageException: java.sql.SQLTransientConnectionException: HikariPool-1 - Connection is not available, request timed out after 72636ms.
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run$$$capture(CompletableFuture.java:1807)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1796)
	at java.base/java.util.concurrent.ForkJoinTask.doExec$$$capture(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
Caused by: org.jobrunr.storage.StorageException: java.sql.SQLTransientConnectionException: HikariPool-1 - Connection is not available, request timed out after 72636ms.
[...]

PR #1492, opened 2026-02-20
Fix AbstractStorageProvider, don't use ForkJoinPool (#1491)

keldkemp:


core/src/main/java/org/jobrunr/storage/AbstractStorageProvider.java
```diff
@@ -43,6 +47,7 @@ protected AbstractStorageProvider(RateLimiter changeListenerNotificationRateLimi
         this.changeListenerNotificationRateLimit = changeListenerNotificationRateLimit;
         this.timerReentrantLock = new ReentrantLock();
         this.notifyJobStatsChangeListenersReentrantLock = new ReentrantLock();
+        this.notificationExecutor = Executors.newSingleThreadExecutor();
```
keldkemp: I'm not sure here. Maybe it's worth doing as in ForkJoinPool - Runtime.getRuntime().availableProcessors() - 1?

auloin: We only need a single thread from what I can tell but this defeats the purpose of `notifyJobStatsChangeListenersReentrantLock`. The only potential issue, is the amount of tasks that may be queued... We may want to look into a discarding thread pool: https://www.baeldung.com/java-rejectedexecutionhandler#3-discard-policy. What do you think @rdehuyss?

core/src/main/java/org/jobrunr/storage/AbstractStorageProvider.java
```diff
@@ -68,6 +73,15 @@ public void removeJobStorageOnChangeListener(StorageProviderChangeListener liste
...
         onChangeListeners.clear();
+        notificationExecutor.shutdown();
+        try {
+            if (!notificationExecutor.awaitTermination(2, TimeUnit.SECONDS)) {
+                notificationExecutor.shutdownNow();
+            }
+        } catch (InterruptedException e) {
+            notificationExecutor.shutdownNow();
+            Thread.currentThread().interrupt();
+        }
```
auloin: I don't think the notification is critical, should we call shutdownNow() directly?

keldkemp: Agreed, we can simplify this part

core/src/main/java/org/jobrunr/storage/AbstractStorageProvider.java
keldkemp: I've looked through the code again. It appears this isn't a particularly important area. I agree we should implement the discard policy

rdehuyss:
Hi @keldkemp - instead of having both a timer and an ExecutorService, I would rather like to move to a `ScheduledThreadPoolExecutor` with a DiscardPolicy. What do you think of this idea? Then we could remove the timer altoghether and it would make the code simpler.

keldkemp:
> Hi @keldkemp - instead of having both a timer and an ExecutorService, I would rather like to move to a `ScheduledThreadPoolExecutor` with a DiscardPolicy. What do you think of this idea? Then we could remove the timer altoghether and it would make the code simpler.

Hi @rdehuyss, I will try to take a look this week and provide the modified code.

core/src/main/java/org/jobrunr/storage/AbstractStorageProvider.java
```diff
@@ -93,33 +106,36 @@ public final void validateRecurringJobInterval(Duration durationBetweenRecurring
...
 
     protected void notifyJobStatsOnChangeListenersIf(boolean mustNotify) {
-        if (mustNotify) {
-            notifyJobStatsOnChangeListeners();
-        }
+        if (!mustNotify) return;
+
+        if (!jobStatsNotificationQueued.compareAndSet(false, true)) return;
+
+        scheduler.execute(() -> {
```
keldkemp: Hi @rdehuyss. I'm not sure about this part.

First — is it really important for us to send the notification "immediately" when we have a recurring task every 5 seconds?
Second — do we need a check here to see if the task is already in progress?

rdehuyss:
Hi @keldkemp - Nice work! I've made some small changes as I think that if there are no listeners, we also don't need to occupy a thread.

rdehuyss:
Before we can merge, we need you to approve our CLA though. If you add yourself now, we can still add this to JobRunr v8.5.

keldkemp:
Thanks) 
I created a PR: https://github.com/jobrunr/clabot-config/pull/69

@rdehuyss

rdehuyss:
@cla-bot check

rdehuyss:
Hi @keldkemp - I'll have a look at build

auloin (approved):

PR #1505, opened 2026-03-02
More sonarqube improvements.

rdehuyss:


core/src/main/java/org/jobrunr/scheduling/ScheduleExpressionType.java
```diff
@@ -32,7 +32,7 @@ public static Schedule createScheduleFromString(String scheduleExpression) {
         if (scheduleExpression.toUpperCase().startsWith("P")) {
             return INTERVAL.createSchedule(scheduleExpression);
         }
-        if (scheduleExpression.matches(".*\\s.*")) {
+        if (scheduleExpression.trim().matches("^(\\S+(?:\\s+\\S+){4,5})(?:\\s+\\[([\\w/]+)])?$")) {
```
auloin, line 35: For consistency: should we re-use the same regex as in JobRunr Pro?

core/src/test/java/org/jobrunr/scheduling/cron/CronExpressionTest.java
```diff
@@ -34,7 +35,8 @@ class CronExpressionTest {
     void testCron(String cronExpression, String baseDate, String expectedResult) {
         try {
             Instant inputInstant = LocalDateTime.parse(baseDate, dateTimeFormatter).toInstant(UTC);
-            CronExpression cron = new CronExpression(cronExpression);
+            CronExpression cron = (CronExpression) ScheduleExpressionType.createScheduleFromString(cronExpression);
+            //CronExpression cron = new CronExpression(cronExpression);
```
auloin: Do we want to keep this?

core/src/main/java/org/jobrunr/scheduling/ScheduleExpressionType.java
rdehuyss: I think the one in OSS is safer due to less backtracking possibilities. So I would try to replace the Pro one with this new OSS one.

core/src/test/java/org/jobrunr/scheduling/cron/CronExpressionTest.java
rdehuyss: Removed it

PR #1539, opened 2026-04-21
Speed up DatabaseCreator.getAllTableNames

### LINKED ISSUE #1538: [BUG] org.jobrunr.storage.sql.common.DatabaseCreator#getAllTableNames is too slow
### JobRunr Version

8.5.2

### JDK Version

eclipse-temurin 17.0.17

### Your SQL / NoSQL database

SQL (Pg, Oralce)

### What happened?

Hello, we are still trying to migrate from version 5.2.0 to 8.5.2. Now we've encountered a new problem.

In `org.jobrunr.storage.sql.common.DatabaseCreator#getAllTableNames`, it pulls all tables from all schemas because the code looks like this:

```java
ResultSet tables = conn.getMetaData().getTables(catalog, null, "%", null)
```
We have databases where there are 10+ schemas, each with 8000 tables. (I set up a test environment in our infrastructure and had to wait 40 minutes and still didn't finish waiting.)

I propose optimizing this. At least something like this to start:
```java
DatabaseMetaData md = conn.getMetaData();

String pattern = "%jobrunr%";
if (md.storesUpperCaseIdentifiers()) pattern = pattern.toUpperCase();
else if (md.storesLowerCaseIdentifiers()) pattern = pattern.toLowerCase();
ResultSet tables = md.getTables(catalog, null, pattern , null)
```
In my case, this gave a huge improvement. It went from over 40 minutes to less than 5 seconds.

Also, I don't quite understand why the current schema isn't used by default? Or was the idea that this is left up to the user and they always need to specify JobRunrProperties.database.tablePrefix?

### How to reproduce?

.

### Relevant log output

```shell

```

keldkemp:
Fixes #1538


core/src/main/java/org/jobrunr/storage/sql/common/DatabaseCreator.java
```diff
@@ -144,7 +145,12 @@ List<String> getAllTableNames() {
         try (final Connection conn = getConnection()) {
             List<String> allTableNames = new ArrayList<>();
             String catalog = conn.getCatalog();
-            try (ResultSet tables = conn.getMetaData().getTables(catalog, null, "%", null)) {
+            DatabaseMetaData md = conn.getMetaData();
+
+            String pattern = "%jobrunr%";
+            if (md.storesUpperCaseIdentifiers()) pattern = pattern.toUpperCase();
+            else if (md.storesLowerCaseIdentifiers()) pattern = pattern.toLowerCase();
+            try (ResultSet tables = md.getTables(catalog, null, pattern, null)) {
```
rdehuyss, line 154: To be sure we're 100% backwards compatible (because we're scared of this change because of all the different DB's and drivers out there. Apparently, this is hard coded in some DB drivers and the MySQL Driver is even funkier. For LowerCase, it asks the database; for uppercase it is always false):

```java
            String pattern = "%";
            if (md.storesMixedCaseIdentifiers()) pattern = "%";
            else if (md.storesUpperCaseIdentifiers()) pattern = "%JOBRUNR_%";
            else if (md.storesLowerCaseIdentifiers()) pattern = "%jobrunr_%";
```

keldkemp:
Hi, I've looked at these changes on our Oracle, MSSQL, and PostgreSQL databases. Seems like everything works as it should.

auloin (approved):

PR #1549, opened 2026-05-04
Improve recurring jobs performance

rdehuyss:


auloin (approved):

patkujawa-wf:
@rdehuyss we just upgraded to 8.6.0 last night and have encountered significant job queuing/backup, wondering if it might have been from this change. We're gathering info as we roll-back and revert, will try to report more thoroughly later but FYI.

PR #1557, opened 2026-05-22
Fix multiple master nodes

rdehuyss:


core/src/main/java/org/jobrunr/server/threadpool/PlatformThreadPoolJobRunrExecutor.java
```diff
@@ -18,17 +26,41 @@ public PlatformThreadPoolJobRunrExecutor(int corePoolSize, String threadNamePref
...
+
+        for (Runnable runnable : toCancel) {
+            ScheduledFuture<?> future = scheduledFutures.remove(runnable);
+            future.cancel(true);
+            executorService.remove(runnable);
+        }
+
+        executorService.purge();
+        executorService.setCorePoolSize(executorService.getCorePoolSize() - toCancel.size());
+        executorService.setMaximumPoolSize(executorService.getMaximumPoolSize() - toCancel.size());
```
auloin: Should we guard against the potential of this setting the pool size to 0?

auloin (approved):

Issue #1569, opened 2026-06-12
[BUG] SQL storage fails on TiDB when getJobsToProcess selects only jobAsJson with ORDER BY / FOR UPDATE SKIP LOCKED


anton-gavrilin:
### JobRunr Version

8.6.1

### JDK Version

25

### Your SQL / NoSQL database

TiDB

## Describe the bug

JobRunr `8.6.1` fails to onboard enqueued jobs when using TiDB through the MySQL JDBC driver.

The background job server repeatedly throws:

```text
org.jobrunr.storage.StorageException: java.sql.SQLException:
Can't find column agi_result_processor.jobrunr_jobs.id in schema
Column: [
  agi_result_processor.jobrunr_jobs.jobasjson,
  agi_result_processor.jobrunr_jobs.state,
  agi_result_processor.jobrunr_jobs.updatedat
]
Unique key: []
```

### How to reproduce?

The failure occurs while OnboardNewWorkTask calls:

DefaultSqlStorageProvider.getJobsToProcess(...)
JobTable.selectJobsToProcess(...)

### Possible fix

Include the primary key in the projection for job-selection queries that use ordering, limits, and row locking:

```
public List<Job> selectJobsToProcess(AmountRequest amountRequest) {
    return withState(ENQUEUED)
           .selectJobs(
                   "id, jobAsJson from jobrunr_jobs where state = :state",
                    pageRequestMapper.map(amountRequest)
                            + dialect.selectForUpdateSkipLocked())
            .collect(toList());
}
```

### Relevant log output

```shell

```

auloin:
Hi @anton-gavrilin,

We don't support TiDB (no tests covering it) so this is more a discussion than a bug. And TiDB doesn't support the `SKIP LOCKED` syntax which we rely on.

In JobRunr Pro, you can choose between optimistic locking or `FOR UPDATE SKIP LOCKED` via configuration. In JobRunr OSS, you'll have to do a bit more work, i.e., wire in a `MySqlStorageProvider` with a dialect that does not use SKIP LOCKED.

anton-gavrilin:
Hi @auloin, what you think about such solution then, will it help for now?

```
    @Bean(name = "storageProvider", destroyMethod = "close")
    StorageProvider storageProvider(DataSource dataSource, JobMapper jobMapper, JobRunrProperties properties) {
[...]

PR #1574, opened 2026-06-17
Add TiDB support to SQL storage provider

isabek:
- Use existing MySQL tests for TiDB

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -154,6 +154,7 @@ dependencies {
     testImplementation 'org.testcontainers:testcontainers-mariadb'
     testImplementation 'org.testcontainers:testcontainers-mssqlserver'
     testImplementation 'org.testcontainers:testcontainers-mysql'
+    testImplementation 'org.testcontainers:testcontainers-tidb'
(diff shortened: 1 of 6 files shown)


anton-gavrilin:
Related raised issue: https://github.com/jobrunr/jobrunr/issues/1569

anton-gavrilin (approved):

MehmetSecgin (approved):

auloin:
Nice job everyone. Happy to see you got it working with minimal effort. We won't add this for now for several reasons:

1. It's easy for the interested users to get it working without the builtin support
2. There isn't enough demand to justify the maintenance burden (especially we don't know how things will turn out in JobRunr Pro and no customer ever asked for TiDB support).

We might reconsider this if there is more demand for TiDB support, till then this PR and the discussion in https://github.com/jobrunr/jobrunr/issues/1569 can serve as guidance on how to build a custom TiDBStorageProvider. We can also add the support if you'd take a JobRunr Pro subscription, the same way we did for CockroachDB.

PR #1577, opened 2026-06-23
Fix Cron expression parsing during DST overlap

rdehuyss:


core/src/main/java/org/jobrunr/scheduling/cron/CronExpression.java
```diff
@@ -357,4 +363,26 @@ private BitSet getUpdatedDays(int year, int month) {
...
+
+    private boolean isDSTOverlapTransition(LocalDateTime candidateLocalDateTime, ZoneId zoneId) {
+        ZoneOffsetTransition transition = zoneId.getRules().getTransition(candidateLocalDateTime);
+        return transition != null && transition.isOverlap();
+    }
+
+    private Instant nextRunDuringDSTOverlap(Instant createdAtInstant, LocalDateTime candidateLocalDateTime, ZoneId zoneId) {
+        ZoneOffsetTransition transition = zoneId.getRules().getTransition(candidateLocalDateTime);
+        Instant justBeforeEndOfOverlap = transition.getDateTimeBefore().atZone(zoneId).toInstant().minusNanos(1);
+        return next(createdAtInstant, justBeforeEndOfOverlap, zoneId);
```
auloin: I think it's better for this method to return `justBeforeEndOfOverlap`. The caller should be responsible for calling `next(...)`.

core/src/main/java/org/jobrunr/scheduling/cron/CronExpression.java
```diff
@@ -357,4 +363,26 @@ private BitSet getUpdatedDays(int year, int month) {
...
+    private boolean isDSTGapTransition(LocalDateTime candidateLocalDateTime, ZoneId zoneId) {
+        ZoneOffsetTransition transition = zoneId.getRules().getTransition(candidateLocalDateTime);
+        return transition != null && transition.isGap();
+    }
+
+    private Instant nextRunDuringDSTGap(Instant createdAtInstant, LocalDateTime candidateLocalDateTime, ZoneId zoneId) {
+        ZoneOffsetTransition transition = zoneId.getRules().getTransition(candidateLocalDateTime);
+        Instant justBeforeEndOfOverlap = transition.getDateTimeAfter().atZone(zoneId).toInstant();
+        return next(createdAtInstant, justBeforeEndOfOverlap, zoneId);
+    }
```
auloin: We don't seem to use these two methods. Do you want to keep them?

auloin (approved):

Issue #1241, opened 2025-04-17
Getting the following error :- java.sql.SQLException: View 'jobrunr_jobs_stats' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them


qingquaner:
### JobRunr Version

7.5.0

### JDK Version

1.8.0_333

### Your SQL / NoSQL database

Mysql 8.0.21

### What happened?

2025-04-17 18:21:27.935 WARN  [ForkJoinPool.commonPool-worker-9] {org.jobrunr.storage.AbstractStorageProvider} Line:203 - Error notifying JobStorageChangeListeners
org.jobrunr.storage.StorageException: java.sql.SQLException: View 'iot_ironman_device.jobrunr_jobs_stats' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them
	at org.jobrunr.storage.sql.common.db.SqlSpliterator.init(SqlSpliterator.java:64) ~[jobrunr-7.5.0.jar:7.5.0]
	at org.jobrunr.storage.sql.common.db.SqlSpliterator.tryAdvance(SqlSpliterator.java:35) ~[jobrunr-7.5.0.jar:7.5.0]
	at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:126) ~[?:1.8.0_333]
	at java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:499) ~[?:1.8.0_333]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:486) ~[?:1.8.0_333]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472) ~[?:1.8.0_333]
	at java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:152) ~[?:1.8.0_333]
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:1.8.0_333]
	at java.util.stream.ReferencePipeline.findFirst(ReferencePipeline.java:464) ~[?:1.8.0_333]
	at org.jobrunr.storage.sql.common.JobStatsView.getJobStats(JobStatsView.java:24) ~[jobrunr-7.5.0.jar:7.5.0]
[...]

Issue #828, opened 2023-09-08
Getting the following error :- java.sql.SQLException: View 'jobrunr_jobs_stats' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them


as-sher:
### JobRunr Version

6.1.3

### JDK Version

1.8

### Your SQL / NoSQL database

mysql 8.0

### What happened?

whenever I open the UI for jobrunr I get this error :-Getting the following error :- java.sql.SQLException: View 'jobrunr_jobs_stats' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them

some methods used in the view are missing in the rds :- rollup_group_item, rollup_sum_switcher

### How to reproduce?

whenever I open the UI for jobrunr I get this error :-Getting the following error :- java.sql.SQLException: View 'jobrunr_jobs_stats' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them

### Relevant log output

```shell
java.sql.SQLException: View 'jobrunr_jobs_stats' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them
```

rdehuyss:
For the open-source version, I do not provide support unless you run the latest and greatest JobRunr version (v6.3.0)

And, in your case, the exception is quite clear and is outside of the scope of JobRunr. I'm afraid I cannot help you as my time to work on the OSS version is limited. A JobRunr Pro subscription comes with support where I can help you.

PR #1590, opened 2026-08-03
Create jobrunr_jobs_stats view with SQL SECURITY INVOKER on MySQL and MariaDB

ThomasVerhoeven1998:
## The issue

On MySQL and MariaDB, a view is created with `SQL SECURITY DEFINER` and `DEFINER = CURRENT_USER` by default. JobRunr's migrations therefore create `jobrunr_jobs_stats` with the database user that happened to run the migrations baked into the view definition.

That `DEFINER` is part of the view definition and is carried over in a database dump. As soon as such a dump is restored into another environment (e.g. production → staging), the `DEFINER` points to a database user that does not exist in the target environment, and every query on the view fails:

```
java.sql.SQLException: View 'jobrunr_jobs_stats' references invalid table(s) or column(s) or function(s)
or definer/invoker of view lack rights to use them
```

```
org.jobrunr.storage.StorageException: ...
	at org.jobrunr.storage.sql.common.JobStatsView.getJobStats(JobStatsView.java:24)
	at org.jobrunr.storage.sql.common.DefaultSqlStorageProvider.getJobStats(...)
```

The result is a JobRunr dashboard that shows no numbers at all, and the only workaround today is to manually run a `CREATE OR REPLACE ... DEFINER=<user valid in the target environment> ... VIEW jobrunr_jobs_stats ...` after every restore.

Related reports: #1241, #828.

## The fix

Recreate `jobrunr_jobs_stats` with `SQL SECURITY INVOKER` on MySQL and MariaDB. The view is then evaluated with the privileges of the user querying it, so the `DEFINER` recorded in the view no longer has to exist in the target environment and a cross-environment restore keeps working.

Verified against MySQL 8 and MariaDB 11 with a `DEFINER` pointing to a non-existent user:

| SQL SECURITY | result |
| --- | --- |
| `DEFINER` (current behaviour) | `ERROR 1045 (28000): Access denied for user ...` |
| `INVOKER` (this PR) | returns the job stats |

### Changes

New migration `v017__change_job_stats_view_to_sql_security_invoker.sql`, following the existing pattern of an empty common migration plus database-specific ones so the numbering stays aligned across all databases:

- `common/migrations/` — empty migration
- `mysql/migrations/` and `mariadb/migrations/` — drop and recreate `jobrunr_jobs_stats` using `CREATE SQL SECURITY INVOKER VIEW`. The view definition itself is unchanged compared to `v016`.

Because it is a regular migration, this fixes both fresh installs and existing databases on upgrade, and it works with a configured table prefix.

### Tests

- `SqlTestUtils.assertJobStatsViewUsesSqlSecurityInvoker(...)` asserts, via `information_schema.views`, that every `jobrunr_jobs_stats` view is created with `SECURITY_TYPE = INVOKER`.
- Called from `AbstractMySQLStorageProviderTest` and `AbstractMariaDbStorageProviderTest`, so it covers `MySQLStorageProviderTest`, `MySQLTablePrefixStorageProviderTest`, `MariaDbStorageProviderTest` and `MariaDbTablePrefixStorageProviderTest`.
- The test was confirmed to fail when the new migration is removed and to pass with it.
- `SqlScriptsTest`, `DatabaseMigrationsProviderTest`, `DatabaseCreatorTest` and `DatabaseCreatorTablePrefixTest` are green.

### Note on privileges

With `SQL SECURITY INVOKER`, the user querying the view needs `SELECT` privileges on `jobrunr_jobs`, `jobrunr_metadata`, `jobrunr_backgroundjobservers` and `jobrunr_recurring_jobs`. The JobRunr application user always has these. A separate, dashboard-only user that was granted rights on the view alone would need those grants added.


ThomasVerhoeven1998:
> We require contributors to sign our Contributor License Agreement, and we don't have @ThomasVerhoeven1998 on file. In order for us to review and merge your code, please create a PR where you add yourself to the [contributors](https://github.com/jobrunr/clabot-config/blob/main/.clabot) of JobRunr. This only needs to be done once. As soon as that is done, we can review your PR.
> 
> Thanks a lot!

https://github.com/jobrunr/clabot-config/pull/84

auloin:
Hi @ThomasVerhoeven1998, I believe we have discussed this in the past. Does the workarounds not work, namely [removing the definer from the dump](https://stackoverflow.com/a/24613430) or using [MySQL Shell Dump & Load utility](https://dev.mysql.com/doc/mysql-shell/8.4/en/mysql-shell-utilities-dump-instance-schema.html) with `strip_definers` to strip it?

ThomasVerhoeven1998:
> Hi @ThomasVerhoeven1998, I believe we have discussed this in the past. Does the workarounds not work, namely [removing the definer from the dump](https://stackoverflow.com/a/24613430) or using [MySQL Shell Dump & Load utility](https://dev.mysql.com/doc/mysql-shell/8.4/en/mysql-shell-utilities-dump-instance-schema.html) with `strip_definers` to strip it?

We have a workaround to recreate the view after every restore which requires manual intervention each time so we are looking for a long term solution. We are using [Percona xtrabackup](http://docs.percona.com/percona-xtrabackup/2.4/index.html) to restore on Percona cluster, so no MySQL tooling. Since Percona xtrabackup is physical backup tool, not a logical dump tool, it can not strip definers

I believe change the view to INVOKER has no negative side effects, that's why a prepped a PR with the changes

Can you elaborate on why you guys prefer DEFINER? And what the benefits are?

auloin:
> Can you elaborate on why you guys prefer DEFINER? And what the benefits are?

Just being cautious. The change is small but it could also be breaking or non desirable for other users.

core/src/main/resources/org/jobrunr/storage/sql/mariadb/migrations/v017__change_job_stats_view_to_sql_security_invoker.sql
```diff
@@ -0,0 +1,22 @@
+DROP VIEW jobrunr_jobs_stats;
+CREATE SQL SECURITY INVOKER VIEW jobrunr_jobs_stats
```
auloin:
```suggestion
CREATE OR REPLACE SQL SECURITY INVOKER VIEW jobrunr_jobs_stats
```

core/src/main/resources/org/jobrunr/storage/sql/mysql/migrations/v017__change_job_stats_view_to_sql_security_invoker.sql
```diff
@@ -0,0 +1,22 @@
+DROP VIEW jobrunr_jobs_stats;
+CREATE SQL SECURITY INVOKER VIEW jobrunr_jobs_stats
```
auloin:
```suggestion
CREATE OR REPLACE SQL SECURITY INVOKER VIEW jobrunr_jobs_stats
```

core/src/test/java/org/jobrunr/storage/sql/SqlTestUtils.java
```diff
@@ -76,4 +82,21 @@ public static void doInTransaction(DataSource dataSource, Exceptions.ThrowingCon
             throw e;
         }
     }
+
+    public static void assertJobStatsViewUsesSqlSecurityInvoker(DataSource dataSource) throws SQLException {
```
auloin: I think it'd be preferable to duplicate this (the method doesn't seem to be portable).

Another possibility is a test in `SqlScriptsTest` to ensure we don't drop the security modifier if the view is changed again for mysql and mariadb.

ThomasVerhoeven1998:
> > Can you elaborate on why you guys prefer DEFINER? And what the benefits are?
> 
> Just being cautious. The change is small but it could also be breaking or non desirable for other users.

Completely understandable, no worries!

auloin:
@cla-bot check

auloin (commented):
The changes look good to me. I just have the below few suggestions.

Could you also change the target to be `v9` branch?

core/src/main/resources/org/jobrunr/storage/sql/mariadb/migrations/v017__change_job_stats_view_to_sql_security_invoker.sql
ThomasVerhoeven1998: Applied, thanks. Verified that `CREATE OR REPLACE` does flip the security type on an already existing view (MySQL 8.0.32 and MariaDB 12.3): starting from the current `v016` view with `SECURITY_TYPE = DEFINER`, after running this migration `information_schema.views` reports `INVOKER`.

It also works when the existing view still has the unusable definer from another environment, so the migration repairs an already broken installation rather than erroring out on it.

core/src/main/resources/org/jobrunr/storage/sql/mysql/migrations/v017__change_job_stats_view_to_sql_security_invoker.sql
ThomasVerhoeven1998: Applied, thanks. Verified that `CREATE OR REPLACE` does flip the security type on an already existing view (MySQL 8.0.32 and MariaDB 12.3): starting from the current `v016` view with `SECURITY_TYPE = DEFINER`, after running this migration `information_schema.views` reports `INVOKER`.

It also works when the existing view still has the unusable definer from another environment, so the migration repairs an already broken installation rather than erroring out on it.

core/src/test/java/org/jobrunr/storage/sql/SqlTestUtils.java
ThomasVerhoeven1998: Agreed, it did not belong in the shared utils. `SqlTestUtils` is back to its original state and the assertion now lives in `AbstractMySQLStorageProviderTest` and `AbstractMariaDbStorageProviderTest`.

I also added your second suggestion, since it guards a different failure mode: `SqlScriptsTest.validateJobStatsViewIsCreatedUsingSqlSecurityInvokerForMySqlAndMariaDb` resolves the effective migration set per storage provider, takes the last statement that creates `jobrunr_jobs_stats` and asserts it contains `SQL SECURITY INVOKER`. So if the view is ever changed again for MySQL/MariaDB without the modifier, that test fails without needing a container. I confirmed it fails when the new migration is removed.

Happy to drop either of the two if you consider it redundant.

ThomasVerhoeven1998:
Thanks for the review @auloin! All three suggestions are applied and the PR is now targeting `v9`.

- Both migrations use a single `CREATE OR REPLACE SQL SECURITY INVOKER VIEW` statement instead of `DROP VIEW` + `CREATE VIEW`.
- `SqlTestUtils` is untouched again; the assertion is duplicated into the MySQL and MariaDB test classes.
- Added a `SqlScriptsTest` guard so the security modifier cannot silently be dropped if the view is changed again.

Apart from the `CREATE` line, the view definition in `v017` is byte-identical to `v016` (including its `cASt` quirk, which I deliberately left alone so the diff stays reviewable).

Verified on MySQL 8.0.32 and MariaDB 12.3 that the migration flips `SECURITY_TYPE` from `DEFINER` to `INVOKER` on an existing view and that it also repairs a view whose definer no longer exists. `MySQLStorageProviderTest`, `MySQLTablePrefixStorageProviderTest`, `MariaDbStorageProviderTest`, `MariaDbTablePrefixStorageProviderTest`, `SqlScriptsTest`, `DatabaseCreatorTest`, `DatabaseCreatorTablePrefixTest` and `DatabaseMigrationsProviderTest` are green, and all modules compile.

On the "could be non desirable for other users" concern, the one behavioural difference worth calling out: with `SQL SECURITY INVOKER` the user querying the view needs `SELECT` on `jobrunr_jobs`, `jobrunr_metadata`, `jobrunr_backgroundjobservers` and `jobrunr_recurring_jobs`. The JobRunr application user always has those, since it reads and writes those tables anyway. The only setup that would regress is one where a separate dashboard-only user was granted rights on the view alone and deliberately denied access to the underlying tables. If you would rather not risk that at all, an alternative is to keep `DEFINER` as the default and make the security clause configurable, but that adds a knob for what I think is a rare setup. Happy to go that route if you prefer.

auloin (approved):

rdehuyss (approved):

PR #1603, opened 2026-08-28
Add jobs filter panel

RomanJobRunr:
TODO: Setup receiving endpoint in api.jobrunr.io

### DIFF
--- a/core/src/main/resources/org/jobrunr/dashboard/frontend/src/utils/helper-functions.js
+++ b/core/src/main/resources/org/jobrunr/dashboard/frontend/src/utils/helper-functions.js
@@ -67,6 +67,13 @@ export function convertToBrowserDefaultDateStyle(date) {
     return dateString.toString().substring(0, dateString.indexOf(' ('));
 }
 
+export function humanReadableNumber(num) {
+    if (typeof num !== 'number' || isNaN(num)) {
+        return '?';
+    }
+    return num.toLocaleString("en", {notation: "compact", compactDisplay: "short"});
+}
+
 export function stringToColor(text) {
     let hash = 0;
     for (let i = 0; i < text.length; i++) {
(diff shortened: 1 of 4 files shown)


rdehuyss:
Needs to go via api.jobrunr.io

RomanJobRunr:
@rdehuyss What would the URL need to be with that change?

core/src/main/resources/org/jobrunr/dashboard/frontend/src/utils/helper-functions.js
```diff
@@ -67,6 +67,13 @@ export function convertToBrowserDefaultDateStyle(date) {
     return dateString.toString().substring(0, dateString.indexOf(' ('));
 }
 
+export function humanReadableNumber(num) {
+    if (typeof num !== 'number' || isNaN(num)) {
+        return '?';
+    }
+    return num.toLocaleString("en", {notation: "compact", compactDisplay: "short"});
+}
+
```
auloin: I don't think this is used in this PR, can it be removed?

auloin (changes requested):

core/src/main/resources/org/jobrunr/dashboard/frontend/src/utils/helper-functions.js
RomanJobRunr: It's used in the dialog title to display the number of jobs they are searching

`jobs-filter-panel.js` line 118 `Search all {total ? (humanReadableNumber(total) + " " + jobState.toLowerCase()) : "your"} jobs in this dashboard`

auloin: Should we just display the raw number there? Btw the `total` displayed there is just the total on the page not on "this dashboard"

RomanJobRunr: The total there is the total for that state true, which is why I word it like "Search all 100 enqueued jobs in this dashboard". I can change it to use the raw number.

auloin (approved):

